PR #646, opened 2023-08-09
maint: Add sonarqube workflow

AlejandroFernandezLuces:
Adds workflow to include SonarQube analysis in the project.


RobPasMue:
Let's close this one for now. If we ever need to come back, we have here your changes

PR #717, opened 2023-09-12
fix: Add missing dependencies

AlejandroFernandezLuces:


jorgepiloto (commented):
I would take advantage of this pull-request to downgrade the versions of some packages to ensure interactive plots are back in the rendered docs.

Here are the required packages:

```console
jupyter==1.0.0
jupytext==1.13.1

nbconvert==6.4.5
notebook==6.4.6
ipython==7.32.0
```

I suggest ignoring the updates for this packages through dependabot configuration, see https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore

The problem with these dependencies is that `jupytext==1.13.1` clashes with `myst-parser==2.0.0` because of the required version of `markdown-it-py`. If I downgrade to `myst-parser==1.0.0`, I would also need to `sphinx<7`, which I think is not acceptable. I'm currently looking at how PyVista is doing this, since they have interactable plots with the latest version of sphinx, still working on it..

AlejandroFernandezLuces:
> I would take advantage of this pull-request to downgrade the versions of some packages to ensure interactive plots are back in the rendered docs.
> 
> Here are the required packages:
> 
> ```
> jupyter==1.0.0
> jupytext==1.13.1
> 
> nbconvert==6.4.5
> notebook==6.4.6
> ipython==7.32.0
> ```
> 
> I suggest ignoring the updates for this packages through dependabot configuration, see https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore

Sorry I edited your comment instead of quoting it 😪 

The problem with these dependencies is that `jupytext==1.13.1` clashes with `myst-parser==2.0.0` because of the required version of `markdown-it-py`. If I downgrade to `myst-parser==1.0.0`, I would also need to `sphinx<7`, which I think is not acceptable. I'm currently looking at how PyVista is doing this, since they have interactable plots with the latest version of sphinx, still working on it..

AlejandroFernandezLuces:
After lots of testing, I've noticed that this is broken in PyVista too for version 0.42 and current dev version. https://docs.pyvista.org/version/dev/index.html


In fact, the current stable documentation is pointing to 0.41, and documentation for 0.42 is missing. I would close this PR, since most likely once this is fixed in PyVista side, documentation plotters will work correctly here. 

I'll open an issue in PyVista about this.

tkoyama010:
Wow! I came here followed the link that was mentioned in the PyVista issue and found this repository is great. I have recently been thinking about creating a library for creating physical model shapes in PyVista and this is the solution for me.

PR #727, opened 2023-09-14
feat: Add measurement widget

### LINKED ISSUE #258: PLOTTING: measure widget
Add feature for measuring the distance between two points within PyVista visualization.

This is dependent upon PyVista implementation. Here is the backlog item:
https://github.com/pyvista/pyvista/issues/588

Once it is merged within PyVista main, can pull it into PyGeometry

AlejandroFernandezLuces:
Closes  #258


RobPasMue (commented):
Is it ready for review? =)

AlejandroFernandezLuces:
> Is it ready for review? =)

Not yet, I'll let you know when it's ready

AlejandroFernandezLuces:
This is the current status of the measurement widget, from what I've seen in https://github.com/pyvista/pyvista/pull/3718/files , the widget is supposed to be snapping to the points of the mesh, but it is not working from PyVista side.

Also, I'm using calls to VTK directly to turn the widget on and off, since using the [PyVista expected method for this](https://docs.pyvista.org/version/stable/api/plotting/_autosummary/pyvista.plotting.widgets.WidgetHelper.clear_measure_widgets.html) seems to be interfering with the picker.


Recording 2023-09-15 at 13 31 08

AlejandroFernandezLuces:
I managed to fix the snapping issue, I had to do way more changes for this to work than I expected. 

Recording 2023-09-15 at 15 59 49

src/ansys/geometry/core/plotting/plotter_helper.py
```diff
@@ -204,6 +226,16 @@ def compute_edge_object_map(self) -> Dict[pv.Actor, EdgePlot]:
             for edge in object.edges:
                 self._edge_actors_map[edge.actor] = edge
 
+    def enable_picking(self):
+        """Enable picking capabilities in the plotter."""
+        self._pl.scene.enable_mesh_picking(
+            callback=self.picker_callback, use_actor=True, show=False, show_message=False
+        )
+
+    def disable_picking(self):
```
RobPasMue, line 235: It seems that enabling picking is activated by default... how do you disable it then?

RobPasMue: I can see the method, but no actual call to it.

AlejandroFernandezLuces: It is called when activating the measurement widget, so it doesn't interfere with each other. 

The function is called here: https://github.com/ansys/pyansys-geometry/blob/c7b52122e06b512c58ed727a13e38e3ca202a0bf/src/ansys/geometry/core/plotting/widgets/measure.py#L71-L74

Regarding the first comment, I forgot to use the `allow_picking` flag to the proper places, already fixed 🙂

src/ansys/geometry/core/plotting/plotter_helper.py
```diff
@@ -68,7 +77,7 @@ def __init__(
         self._picked_list = set()
         self._picker_added_actors_map = {}
         self._edge_actors_map = {}
-
+        self._widgets = []
         if self._use_trame and _HAS_TRAME:
```
RobPasMue, line 81:
```suggestion
        self._widgets = []

        if self._use_trame and _HAS_TRAME:
```

RobPasMue (dismissed):
Great job @AlejandroFernandezLuces - Update the cache and let's merge it.

Unless you prefer to do the increased transparency of the black shapes here.

RobPasMue (commented):
Looks good! Let's bump the cache

RobPasMue (approved):

PR #730, opened 2023-09-15
feat: skip unstable images temporarily

RobPasMue:
As title says


.github/workflows/ci_cd.yml
```diff
@@ -278,16 +282,20 @@ jobs:
     name: Testing and coverage (Linux)
     needs: [smoke-tests]
     runs-on: ubuntu-latest
+    if:
```
RobPasMue, line 285:
```suggestion
```

PR #735, opened 2023-09-18
feat: Add name filtering in the plotter

AlejandroFernandezLuces:
## Overview

Adds the option to filter names in the plotter, by accepting a regular expression to add only the bodies that match the regex.


src/ansys/geometry/core/plotting/plotter.py
```diff
@@ -389,7 +394,11 @@ def add(
         elif isinstance(object, Sketch):
             self.plot_sketch(object, **plotting_options)
         elif isinstance(object, Body) or isinstance(object, MasterBody):
-            self.add_body(object, merge_bodies, **plotting_options)
+            if filter:
```
RobPasMue, line 397: IMO, the filter should apply to everything, not only bodies but also components, and other objects that may come in the future (i.e. beams)

AlejandroFernandezLuces: Changed, now the filter is applied to every object that has `name` attribute.

src/ansys/geometry/core/plotting/plotter.py
```diff
@@ -379,7 +384,11 @@ def add(
             Mapping between the ~pyvista.Actor and the PyAnsys Geometry object.
         """
         logger.debug(f"Adding object type {type(object)} to the PyVista plotter")
-
+        if filter:
+            if hasattr(object, "name"):
+                if not re.search(filter, object.name):
+                    logger.info(f"Name {object.name} not found in regex {filter}.")
```
RobPasMue, line 390: Use debug rather than info. Too verbose otherwise
```suggestion
                    logger.debug(f"Name {object.name} not found in regex {filter}.")
```

src/ansys/geometry/core/plotting/plotter.py
```diff
@@ -379,7 +384,11 @@ def add(
...
+                    return self._geom_object_actors_map
         if isinstance(object, List) and isinstance(object[0], pv.PolyData):
```
RobPasMue, line 392:
```suggestion
                    return self._geom_object_actors_map

        # Check what kind of object we are dealing with
        if isinstance(object, List) and isinstance(object[0], pv.PolyData):
```

RobPasMue (approved):
Some minor comments only. LGTM!

RobPasMue (approved):

PR #747, opened 2023-09-20
fix: Add opacity to plot selection

AlejandroFernandezLuces:
Some plotters still had fully black objects in the scene due to plot selection not being able to separate edges from faces. This PR adds transparency to all objects represented through plot_selection.


src/ansys/geometry/core/sketch/sketch.py
```diff
@@ -850,6 +850,7 @@ def plot(
                 selected_pd_objects,
                 screenshot=screenshot,
                 view_2d=view_2d_dict,
+                opacity=0.7,
```
RobPasMue, line 853:
```suggestion
                opacity=0.7, # Since we are passing PD objects directly... we apply opacity to all of them
```

RobPasMue (approved):

RobPasMue (approved):

RobPasMue:
Yep, go ahead! 😄

RobPasMue (approved):

PR #750, opened 2023-09-21
docs: error on docstrings for launcher methods

RobPasMue:
As title says

### DIFF
--- a/src/ansys/geometry/core/connection/launcher.py
+++ b/src/ansys/geometry/core/connection/launcher.py
@@ -387,7 +387,7 @@ def launch_modeler_with_discovery(
 
         **Ansys products versions and their corresponding int values:**
 
-        * 24.1 -> value : ``241``
+        * ``241`` : Ansys 24R1
     host: str, optional
         IP address at which the Discovery session will be deployed. By default,
         its value will be ``localhost``.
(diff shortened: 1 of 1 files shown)


src/ansys/geometry/core/connection/launcher.py
```diff
@@ -380,10 +380,11 @@ def launch_modeler_with_discovery(
...
         a SystemError will be raised.
-        Possible values:
-        Ansys products Versions and their corresponding int values:
-        23.2 -> value : 232
-        24.1 -> value : 241
+
+        **Ansys products versions and their corresponding int values:**
+
+        * 23.2 -> value : ``232``
+        * 24.1 -> value : ``241``
```
AlejandroFernandezLuces, line 387: It's kind of obvious that it is the int equivalent right? Also, to be uniform with the other lists of equivalences I would put first the int value and then what it represents.
```suggestion
        * ``232`` : 23.2
        * ``241`` : 24.1
```

src/ansys/geometry/core/connection/launcher.py
```diff
@@ -472,10 +474,11 @@ def launch_modeler_with_spaceclaim(
...
```
AlejandroFernandezLuces, line 481:
```suggestion
        * ``232`` : 23.2
        * ``241`` : 24.1
```

AlejandroFernandezLuces (dismissed):
LGTM, a minor comment you might want to consider.

AlejandroFernandezLuces (dismissed):

RobPasMue:
Need re-approval

AlejandroFernandezLuces (approved):

PR #774, opened 2023-09-28
feat: add plane clipping capabilities to plotter

AlejandroFernandezLuces:
Integration test is ready in local, but I want to make sure this is what we are looking for before messing with the image cache.

### DIFF
--- a/src/ansys/geometry/core/plotting/plotter.py
+++ b/src/ansys/geometry/core/plotting/plotter.py
@@ -230,6 +230,12 @@ class Plotter:
         if show_frame:
             self.plot_frame(sketch._plane)
 
+        if "clip" in plotting_options:
+            logger.warning("Clipping is not available in Sketch objects.")
+            plotting_options.pop("clip")
+            plotting_options.pop("normal", None)
+            plotting_options.pop("origin", None)
+
         self.add_sketch_polydata(sketch.sketch_polydata_faces(), opacity=0.7, **plotting_options)
         self.add_sketch_polydata(sketch.sketch_polydata_edges(), **plotting_options)
 
@@ -280,6 +286,11 @@ class Plotter:
         # Use the default PyAnsys Geometry add_mesh arguments
         self.__set_add_mesh_defaults(plotting_options)
         dataset = body.tessellate(merge=merge)
+        if "clip" in plotting_options:
+            plotting_options.pop("clip")
+            dataset = self.clip(dataset, **plotting_options)
+            plotting_options.pop("normal", None)
+            plotting_options.pop("origin", None)
         if isinstance(dataset, pv.MultiBlock):
             actor, _ = self.scene.add_composite(dataset, **plotting_options)
         else:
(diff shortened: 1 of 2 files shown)


RobPasMue:
@rajesh1359 can you confirm this is what you were asking for in terms of clipping?

RobPasMue:
@AlejandroFernandezLuces @rajesh1359 what's the status here? Does this solve our plotting requests?

AlejandroFernandezLuces:
This implementation works for what @rajesh1359 wanted to accomplish. He also suggested that we could implement clipping directly in the plotter, like in https://github.com/ansys-internal/kitware-ansys/issues/16 , but I think we could leave that for another PR.

For the time being, we can wrap up this PR. Can you give this a review @RobPasMue ?

RobPasMue (changes requested):
@AlejandroFernandezLuces - can we implement some test cases in which from a plotted design we perform some clipping operations?

src/ansys/geometry/core/plotting/plotter.py
```diff
@@ -281,6 +287,11 @@ def add_body(
         # Use the default PyAnsys Geometry add_mesh arguments
         self.__set_add_mesh_defaults(plotting_options)
         dataset = body.tessellate(merge=merge)
+        if "clip" in plotting_options:
+            plotting_options.pop("clip")
+            dataset = self.clip(dataset, **plotting_options)
+            plotting_options.pop("normal", None)
+            plotting_options.pop("origin", None)
```
RobPasMue, line 294: Why do you need to pop these options?

src/ansys/geometry/core/plotting/plotter.py
```diff
@@ -344,8 +362,44 @@ def add_sketch_polydata(self, polydata_entries: List[pv.PolyData], **plotting_op
...
+
+        self.scene.add_mesh(mb, color=EDGE_COLOR, **plotting_options)
+
+    def clip(
+        self,
+        mesh: Union[pv.PolyData, pv.MultiBlock],
+        normal: str = "x",
+        origin: tuple = None,
+        **plotting_options,
```
RobPasMue, line 382: Why not pass a plane rather than all these options? It's more intuitive

RobPasMue (changes requested):

src/ansys/geometry/core/plotting/plotter.py
AlejandroFernandezLuces: Not sure what you mean here, these are the parameters needed to pass to PyVista

src/ansys/geometry/core/plotting/plotter.py
AlejandroFernandezLuces: If these are not popped, I'll get errors when passing the plotting_options to `add_mesh` function.

src/ansys/geometry/core/plotting/plotter.py
RobPasMue: All those parameters, you can extract them from our Plane object. It makes much more sense to use a "Plane" rather than the PyVista expected inputs. You are missing units there for example...

RobPasMue (approved):
Looks great! Thanks @AlejandroFernandezLuces

RobPasMue:
We should consider adding an example about this. Could you open an follow up PR with this?

AlejandroFernandezLuces:
> We should consider adding an example about this. Could you open a follow up PR with this?

Good idea! Let's merge this since it is done and I'll take care of the example.

PR #777, opened 2023-09-29
Secure get_available_port method.

b-matteo:
Use socket.gethostname() instead of empty string to get an available port.


RobPasMue (approved):
LGTM! Thanks @b-matteo

PR #779, opened 2023-09-29
Remove staticmethod decorator

rmanno91:
The method get_available_port is not contained in any calls and decorated with staticmethod therefore it is not callable. Removed


RobPasMue (dismissed):
Nice catch @rmanno91!!

RobPasMue (approved):

PR #789, opened 2023-10-04
feat: adding Python 3.12 support

### LINKED ISSUE #788: Add support for Python 3.12
Related to #787 - we tried it but ``vtk`` was blocking the update.

RobPasMue:
Closes #788


PipKat (dismissed):
LGTM.

RobPasMue:
Blocked due to VTK... waiting for fix.

RobPasMue:
Seems like ``9.3.0rc2`` solved our build issues at least. However, we shouldn't allow for dev/pre-releases - and rather use actual releases. Waiting for ``9.3.0`` to come out

pyproject.toml
```diff
@@ -36,7 +37,7 @@ dependencies = [
     "pyvista>=0.37.0",
     "scipy>=1.7.3",
     "six>=1.16.0",
-    "vtk>=9",
+    "vtk>=9.dev0",
```
RobPasMue, line 40:
```suggestion
    "vtk>=9",
```

AlejandroFernandezLuces:
Should we update this line too?

https://github.com/ansys/pyansys-geometry/blob/3c4867597e1712d469b14e37f116574c1f09cfcc/.github/workflows/ci_cd.yml#L12

https://github.com/ansys/pyansys-geometry/blob/3c4867597e1712d469b14e37f116574c1f09cfcc/.github/workflows/nightly_docker_test.yml#L8
If that's the case, we should update several comments:

https://github.com/ansys/pyansys-geometry/blob/3c4867597e1712d469b14e37f116574c1f09cfcc/.github/workflows/ci_cd.yml#L92


Also, should this URL be updated?
https://github.com/ansys/pyansys-geometry/blob/3c4867597e1712d469b14e37f116574c1f09cfcc/doc/source/conf.py#L120

RobPasMue:
> Should we update this line too?
> 
> https://github.com/ansys/pyansys-geometry/blob/3c4867597e1712d469b14e37f116574c1f09cfcc/.github/workflows/ci_cd.yml#L12
> 
> https://github.com/ansys/pyansys-geometry/blob/3c4867597e1712d469b14e37f116574c1f09cfcc/.github/workflows/nightly_docker_test.yml#L8
> 

I would still use 3.11 for some time... 3.12 "just" came out. I changed to 3.11 as main version not so long ago. Let's give it some time still.

> 
> Also, should this URL be updated?
> 
> https://github.com/ansys/pyansys-geometry/blob/3c4867597e1712d469b14e37f116574c1f09cfcc/doc/source/conf.py#L120

3.12 docs had some issues with bool types and they are not being linked properly. I would avoid using it for now.

PR #810, opened 2023-10-16
Feat/change design id compression

b-matteo:
In order to handle properly multi documents on backend, the ApiServer now use the compressend encoding method for Design objects instead of ObjectPath. This is because with ObjectPath method, all the Design id had the same value: "0:2". 
With the compressed encoding, each Design has a unique ID. 
This branch aims to adapt PyAnsys-Geometry to those changes on backend's side.


jonahrb (approved):

RobPasMue (changes requested):
We have to distinguish between the root component ID and the Design ID itself now... This changes are not going to solve the issue since we are mixing them

RobPasMue (approved):

PR #815, opened 2023-10-18
feat: add manifest path product launcher

b-matteo:
Adding manifest_path parameters to product (Discovery and SC) local launcher in order to be able to specify a specific version of the apiserver with an other version of the product.
For instance, this allows starting Discovery 2022.1 but using ApiServerAddin 2024.1

Also, fixing the GeometryService folder's name in the unified installer.


src/ansys/geometry/core/connection/product_instance.py
```diff
@@ -267,11 +272,17 @@ def _is_port_available(port: int, host: str = "localhost") -> bool:
...
+def _manifest_path_provider(
+    version: int, available_installations: Dict, manifest_path: str = None
+) -> str:
     """Return the ApiServer's addin manifest file path."""
-    return os.path.join(
-        available_installations[version], ADDINS_SUBFOLDER, BACKEND_SUBFOLDER, MANIFEST_FILENAME
-    )
+    if manifest_path != None and os.path.exists(manifest_path):
+        return manifest_path
+    elif manifest_path != None and os.path.exists(manifest_path) == False:
```
RobPasMue, line 281:
```suggestion
    elif manifest_path != None and not os.path.exists(manifest_path):
```

RobPasMue (changes requested):
Merge my suggestions, otherwise approved

src/ansys/geometry/core/connection/product_instance.py
jonahrb, line 281: I'll go one step further and say 😄:
```suggestion
    if manifest_path is not None and os.path.exists(manifest_path):
        return manifest_path
    elif manifest_path is not None and not os.path.exists(manifest_path):
```

RobPasMue: I can only agree with @jonahrb! =)

src/ansys/geometry/core/connection/product_instance.py
```diff
@@ -35,7 +35,7 @@
     from ansys.geometry.core.modeler import Modeler
 
 
-WINDOWS_GEOMETRY_SERVICE_FOLDER = "GeometryServices"
+WINDOWS_GEOMETRY_SERVICE_FOLDER = "GeometryService"
```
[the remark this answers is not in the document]
b-matteo: I tested it against the last unified installer, it is ideed without the final "s"

src/ansys/geometry/core/connection/product_instance.py
b-matteo: Done =)

RobPasMue (approved):

RobPasMue:
@kmahajan-cadfem - I just remembered that you were the one proposing the change in the folder name  some time ago. Looks like 24R1 will be shipped with the new folder name. Please verify on your end when the time comes 😃

PR #819, opened 2023-10-20
feat: handling multiple designs

### LINKED ISSUE #792: Handle multiple active designs
Multiple active designs should start to be handled now that we have the needed API to activate various designs.

RobPasMue:
Closes #792


tests/integration/test_design.py
```diff
@@ -1684,3 +1685,61 @@ def test_child_component_instances(modeler: Modeler):
...
+
+    # Create a Sketch object and draw a rectangle
+    sketch2 = Sketch()
+    sketch2.box(Point2D([-30, -30], UNITS.mm), 5 * UNITS.mm, 8 * UNITS.mm)
+
+    # Extrude the sketch to create a body
+    design2.extrude_sketch("MyRectangle", sketch2, Quantity(10, UNITS.mm))
+
+    # Asking for the plotting of the first design should lead to an error
+    # since the tessellation of the first design has not been computed yet
```
RobPasMue, line 1724:
```suggestion
    # Initiate expected output images
```

src/ansys/geometry/core/designer/designpoint.py
```diff
@@ -28,6 +28,8 @@
 from ansys.geometry.core.misc.units import UNITS
 
 if TYPE_CHECKING:  # pragma: no cover
+    import pyvista as pv
```
AlejandroFernandezLuces, line 31: If we are going to import PyVista for typechecking, can we import it for the whole module and avoid the lazy import in `_to_polydata` method? What do you think?

src/ansys/geometry/core/misc/checks.py
```diff
@@ -20,10 +20,61 @@
...
+def ensure_design_is_active(method):
+    """
+    Make sure that the design is active before executing a method.
+
+    This function is necessary to be called whenever we do any operation on the design.
+    If we are just accessing information of the class, it is not necessary to call this.
+    """
+
+    def wrapper(self, *args, **kwargs):
+        # Check if the current design is active... otherwise activate it
```
AlejandroFernandezLuces, line 40: Minor comment suggestion
```suggestion
        # Check if the current design is active, otherwise activate it
```

src/ansys/geometry/core/misc/checks.py
```diff
@@ -20,10 +20,61 @@
...
+        def get_design_ref(obj) -> "Design":
+            if hasattr(obj, "_modeler"):  # In case of a Design object
+                return obj
+            elif hasattr(
+                obj, "_parent_component"
+            ):  # In case of a Body, Component, DesignPoint, Beam
+                return get_design_ref(obj._parent_component)
+            elif hasattr(obj, "_body"):  # In case of a Face, Edge
+                return get_design_ref(obj._body._parent_component)
```
AlejandroFernandezLuces, line 49: Can you add a comment here to indicate the recursion? It will make the function a bit easier to read

AlejandroFernandezLuces (changes requested):
Nice work! 🙂 

Since this PR will cause performance issues, we could implement a selector to let users decide whether they need to use multiple designs or not. As discussed offline, a good option would be to create a global flag to perform this selection.

src/ansys/geometry/core/designer/designpoint.py
RobPasMue: Hmm I don't see it. TYPE_CHECKING is only rendered at IDE level. Not when running the code. Which means that it does not really get imported. I would still leave it where it is :)

src/ansys/geometry/core/misc/checks.py
RobPasMue: Yep for sure!

RobPasMue:
> Nice work! 🙂
> 
> Since this PR will cause performance issues, we could implement a selector to let users decide whether they need to use multiple designs or not. As discussed offline, a good option would be to create a global flag to perform this selection.

Agreed. As we discussed, I think it is a great suggestion. Thanks @AlejandroFernandezLuces! By default I would leave it active for multiple designs.

AlejandroFernandezLuces (approved):
LGTM

PR #821, opened 2023-10-23
feat: reactivate PMDB export

### LINKED ISSUE #710: PMDB export is not working on binaries uploaded
This requires further investigation on the binaries uploaded, assigning to @b-matteo. In the meantime, we will comment out this specific part of the test.

RobPasMue:
Closes #710


RobPasMue:
Blocked, just tested using the latest binaries and it is still failing. It now also fails with the uploaded Docker images.

PR #839, opened 2023-11-03
maint: Reactivate style action

AlejandroFernandezLuces:
This PR is to reactivate vale and fix any error that might arise


AlejandroFernandezLuces:
I removed the Google.Headings rule, as it doesn't work as expected in its latest version. See https://github.com/ansys/pyprimemesh/pull/639

RobPasMue (approved):

PR #843, opened 2023-11-07
fix: Revert dependencies due to broken plotters

AlejandroFernandezLuces:


AlejandroFernandezLuces:
Let's revert to static images temporarily while we look for the error, what do you think @RobPasMue ?

image

src/ansys/geometry/core/plotting/plotter.py
```diff
@@ -509,7 +509,7 @@ def show(
         # Conditionally set the Jupyter backend as not all users will be within
         # a notebook environment to avoid a pyvista warning
         if self.scene.notebook and jupyter_backend is None:
-            jupyter_backend = "panel"
+            jupyter_backend = "static"
```
RobPasMue, line 512: Add a TODO here and open an issue

PR #849, opened 2023-11-13
Rotate a body

jonahrb:


tests/integration/test_design.py
```diff
@@ -777,6 +777,23 @@ def test_bodies_translation(modeler: Modeler):
...
+    """Test for verifying the correct rotation of a ``Body``."""
+
+    # Create your design on the server side
+    design = modeler.create_design("BodyRotation_Test")
+
+    body = design.extrude_sketch("box", Sketch().box(Point2D([0, 0]), 1, 1), 1)
+    copy = body.copy(design, "copy")
+    body.rotate(Point3D([0, 0, 0]), UnitVector3D([0, 0, 1]), np.pi / 4)
+
+    # If rotated, uniting body and its copy should now result in a volume > 1
```
RobPasMue, line 790: Is it possible to perform some more checks on the Linux side?

tests/integration/test_design.py
```diff
@@ -777,6 +777,23 @@ def test_bodies_translation(modeler: Modeler):
...
+    # boolean operations not on linux?
+    if modeler.client.backend_type is not BackendType.LINUX_SERVICE:
+        body.unite(copy)
+        assert body.volume.m > 1
+
+
```
RobPasMue, line 796: Nice way of verifying the rotation 😄

src/ansys/geometry/core/designer/body.py
```diff
@@ -740,6 +767,27 @@ def translate(
...
+        rotation_magnitude = angle.value.m_as(DEFAULT_UNITS.SERVER_ANGLE)
+        self._grpc_client.log.debug(f"Rotating body {self.id}.")
+        self._bodies_stub.Rotate(
+            RotateRequest(
+                id=self.id,
+                axis_origin=point3d_to_grpc_point(axis_origin),
+                axis_direction=unit_vector_to_grpc_direction(axis_direction),
+                angle=rotation_magnitude,
+            )
+        )
```
RobPasMue, line 789: Is there no response coming from the server side?

RobPasMue: A true/false message should be enough to see whether the body was effectively rotated I guess.

jonahrb: @RobPasMue `rotate()` is basically identical to `translate()` in function, so I wanted to stay consistent between the 2 in design

tests/integration/test_design.py
jonahrb: @RobPasMue just pushed new changes to the test that are cross-platform :)

RobPasMue (dismissed):

RobPasMue:
Looks like the Linux side of things is failing

jonahrb:
> Looks like the Linux side of things is failing

@RobPasMue yeah not sure what's going on there, looks like it's not importing RotateRequest, which doesn't make sense

RobPasMue (approved):

PR #873, opened 2023-12-04
fix: Add default dict to `run_discovery_script_file` 

### LINKED ISSUE #870: ``script_args`` should be an optional argument on ``run_discovery_script_file`` argument
### 🔍 Before submitting the issue

- [X] I have searched among the existing issues
- [X] I am using a Python virtual environment

### 🐞 Description of the bug

I want to run a script in SpaceClaim using run_discovery_script_file, but I'm unsure what to include in script_args. I have reviewed the original documentation, but it did not provide guidance on how to manage this.

from ansys.geometry.core import launch_modeler_with_spaceclaim

modeler_discovery = launch_modeler_with_spaceclaim()

modeler_discovery.run_discovery_script_file(file_path=r'E:\2023project\nncx\muti-optimize\Mesh_Gem\1.scscript',script_args=script_args)
image

### 📝 Steps to reproduce

I want to run a script in SpaceClaim using run_discovery_script_file, but I'm unsure what to include in script_args. I have reviewed the original documentation, but it did not provide guidance on how to manage this.

### 💻 Which operating system are you using?

Windows

### 📀 Which ANSYS version are you using?

23r2

### 🐍 Which Python version are you using?

3.10

### 📦 Installed packages

```shell
numpy
pandas
ansys-geometry...
```

AlejandroFernandezLuces:
Defaults `script_args` to `{}`.

Closes  #870

### DIFF
--- a/src/ansys/geometry/core/modeler.py
+++ b/src/ansys/geometry/core/modeler.py
@@ -304,7 +304,7 @@ class Modeler:
 
     @protect_grpc
     def run_discovery_script_file(
-        self, file_path: str, script_args: Dict[str, str], import_design=False
+        self, file_path: str, script_args: Dict[str, str] = {}, import_design=False
     ) -> Tuple[Dict[str, str], Optional["Design"]]:
         """
         Run a Discovery script file.
--- a/src/ansys/geometry/core/modeler.py
+++ b/src/ansys/geometry/core/modeler.py
@@ -319,13 +319,13 @@ class Modeler:
         file_path : str
             Path of the file. The extension of the file must be included.
         script_args : Optional[Dict[str, str]], optional.
-            Arguments to pass to the script. By default, ``None``
+            Arguments to pass to the script. By default, ``None``.
         import_design : bool, optional.
             Whether to refresh the current design from the service. When the script
             is expected to modify the existing design, set this to ``True`` to retrieve
             up-to-date design data. When this is set to ``False`` (default) and the
             script modifies the current design, the design may be out-of-sync. By default,
-            ``False``
+            ``False``.
 
         Returns
         -------
--- a/src/ansys/geometry/core/modeler.py
+++ b/src/ansys/geometry/core/modeler.py
@@ -304,7 +304,7 @@ class Modeler:
 
     @protect_grpc
     def run_discovery_script_file(
-        self, file_path: str, script_args: Dict[str, str] = None, import_design=False
+        self, file_path: str, script_args: Optional[Dict[str, str]] = None, import_design=False
     ) -> Tuple[Dict[str, str], Optional["Design"]]:
         """
         Run a Discovery script file.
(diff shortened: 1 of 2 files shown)


src/ansys/geometry/core/modeler.py
```diff
@@ -304,7 +304,7 @@ def __repr__(self) -> str:
 
     @protect_grpc
     def run_discovery_script_file(
-        self, file_path: str, script_args: Dict[str, str], import_design=False
+        self, file_path: str, script_args: Dict[str, str] = {}, import_design=False
```
RobPasMue, line 307: Mmm.. not a fan of this. Default it to ``None``, add the ``Optional`` typehint and actively check in the code whether it is None or not. Do something like

```py
script_args = script_args if script_args else {}
```

Reason is that setting an empty dictionary as default value may lead to undesired behavior --> https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python

I know we are not editing the dictionary inside, but it is just best to avoid these issues.

RobPasMue (changes requested):

src/ansys/geometry/core/modeler.py
```diff
@@ -304,7 +304,7 @@ def __repr__(self) -> str:
 
     @protect_grpc
     def run_discovery_script_file(
-        self, file_path: str, script_args: Dict[str, str], import_design=False
+        self, file_path: str, script_args: Dict[str, str] = None, import_design=False
```
RobPasMue, line 307: @AlejandroFernandezLuces - adapt docstring to state that is is optional

RobPasMue: Add the typehint as well.

RobPasMue (changes requested):

src/ansys/geometry/core/modeler.py
```diff
@@ -318,13 +318,13 @@ def run_discovery_script_file(
         ----------
         file_path : str
             Path of the file. The extension of the file must be included.
-        script_args : dict[str, str]
-            Arguments to pass to the script.
-        import_design : bool, default: False
+        script_args : dict[str, str], optional.
+            Arguments to pass to the script. By default, `None`
```
RobPasMue, line 322:
```suggestion
            Arguments to pass to the script. By default, ``None``
```
Align with library standards... we are using double ticks instead of single ticks...

src/ansys/geometry/core/modeler.py
```diff
@@ -318,13 +318,13 @@ def run_discovery_script_file(
         ----------
         file_path : str
             Path of the file. The extension of the file must be included.
-        script_args : dict[str, str]
-            Arguments to pass to the script.
-        import_design : bool, default: False
+        script_args : dict[str, str], optional.
```
RobPasMue, line 321:
```suggestion
        script_args : Optional[Dict[str, str]], optional.
```

src/ansys/geometry/core/modeler.py
```diff
@@ -318,13 +318,13 @@ def run_discovery_script_file(
...
+            Arguments to pass to the script. By default, `None`
+        import_design : bool, optional.
             Whether to refresh the current design from the service. When the script
             is expected to modify the existing design, set this to ``True`` to retrieve
             up-to-date design data. When this is set to ``False`` (default) and the
-            script modifies the current design, the design may be out-of-sync.
+            script modifies the current design, the design may be out-of-sync. By default, `False`
```
RobPasMue, line 327:
```suggestion
            script modifies the current design, the design may be out-of-sync. By default, ``False``.
```

RobPasMue (changes requested):

PR #896, opened 2023-12-26
Add an installation section to the getting started.

MaxJPRey:
I thought that it could have an added value to quickly see how to install the package when starting to look at the `getting started`.


RobPasMue (approved):
LGTM! Thanks @MaxJPRey ❤️

PR #905, opened 2024-01-03
fix: Broken edge picking

AlejandroFernandezLuces:
Edge picking was broken due to a change in the object I was checking. Now it is working.

### DIFF
--- a/src/ansys/geometry/core/plotting/plotter_helper.py
+++ b/src/ansys/geometry/core/plotting/plotter_helper.py
@@ -229,10 +229,11 @@ class PlotterHelper:
         """
         for object in self._geom_object_actors_map.values():
             # get edges only from bodies
+            geomObj = object.object
             if (
-                isinstance(object, Body)
-                or isinstance(object, MasterBody)
-                or isinstance(object, Face)
+                isinstance(geomObj, Body)
+                or isinstance(geomObj, MasterBody)
+                or isinstance(geomObj, Face)
             ):
                 for edge in object.edges:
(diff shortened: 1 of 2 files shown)


RobPasMue:
Looks like some objects are missing when picking

AlejandroFernandezLuces:
I can add them in this PR

src/ansys/geometry/core/plotting/plotter_helper.py
```diff
@@ -229,10 +231,13 @@ def compute_edge_object_map(self) -> Dict[pv.Actor, EdgePlot]:
...
+            geom_obj = object.object
             if (
-                isinstance(object, Body)
-                or isinstance(object, MasterBody)
-                or isinstance(object, Face)
+                isinstance(geom_obj, Body)
+                or isinstance(geom_obj, MasterBody)
+                or isinstance(geom_obj, Face)
+                or isinstance(geom_obj, SketchFace)
+                or isinstance(geom_obj, Sketch)
```
RobPasMue, line 240: Shouldn't it be SketchEdge?

RobPasMue (changes requested):

src/ansys/geometry/core/plotting/plotter_helper.py
AlejandroFernandezLuces: In this part of the code we need to get the objects that have a list of edges as an attribute, not the edges themselves

RobPasMue (approved):

PR #923, opened 2024-01-12
Get active design

umutsoysalansys:
Currently there is no method the get active design other than read_existing_design() which creates a new design object and returns it. With this method we can get the active design on the modeler.

### DIFF
--- a/src/ansys/geometry/core/modeler.py
+++ b/src/ansys/geometry/core/modeler.py
@@ -179,6 +179,8 @@ class Modeler:
             if self._designs[key]._is_active:
                 return self._designs[key]
 
+        return None
+
     def read_existing_design(self) -> "Design":
         """
         Read the existing design on the service with the connected client.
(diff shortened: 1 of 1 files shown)


src/ansys/geometry/core/modeler.py
```diff
@@ -166,6 +166,21 @@ def create_design(self, name: str) -> "Design":
...
+        Get the active design on the modeler object.
+
+        Returns
+        -------
+        Design
+            Design object already existing on the modeler.
+        """
+        for key, _ in self._designs.items():
+            if self._designs[key]._is_active:
+                return self._designs[key]
```
RobPasMue, line 180:
```suggestion
        for _ , design in self._designs.items():
            if design ._is_active:
                return design
```

RobPasMue: No need to just extract the keys and then extract the value given a key... you can loop on the values directly.

RobPasMue (approved):

RobPasMue:
@umutsoysalansys - feel free to merge if you agree with my last commits

b-matteo (approved):
LGTM, now @umutsoysalansys we should use that into the fix method to make sure we have an up to date client after a fix call.

umutsoysalansys:
@RobPasMue  thanks for the suggestions.

PR #933, opened 2024-01-17
active design sync after repair operations,

umutsoysalansys:
Added read existing design to each fix method and you can see the current flow in the caption below. 

image


src/ansys/geometry/core/tools/repair_tools.py
```diff
@@ -33,6 +33,14 @@
 from ansys.api.geometry.v0.repairtools_pb2_grpc import RepairToolsStub
 from beartype.typing import TYPE_CHECKING, List
 
+if TYPE_CHECKING:  # pragma: no cover
+    from ansys.geometry.core.designer.body import Body
+
+from beartype.typing import TYPE_CHECKING, List
+
+if TYPE_CHECKING:  # pragma: no cover
+    from ansys.geometry.core.modeler import Modeler
```
RobPasMue, line 42: Put them into a single if @umutsoysalansys

src/ansys/geometry/core/tools/problem_areas.py
```diff
@@ -242,12 +258,15 @@ class SmallFaceProblemAreas(ProblemArea):
         Active supporting geometry service instance for design modeling.
     faces : List[str]
         List of edges associated with the design.
+    modeler : Modeler
+        modeler object
```
RobPasMue, line 262: Please format properly docstrings. No typos.
```suggestion
        Modeler object.
```

src/ansys/geometry/core/tools/problem_areas.py
```diff
@@ -282,16 +302,19 @@ class SplitEdgeProblemAreas(ProblemArea):
...
     id : str
         Server-defined ID for the body.
-    grpc_client : GrpcClient
-        Active supporting geometry service instance for design modeling.
     edges : List[str]
         List of edges associated with the design.
+    grpc_client : GrpcClient
+        Active supporting geometry service instance for design modeling.
+    modeler : Modeler
+        modeler object
```
RobPasMue, line 310:
```suggestion
        Modeler object.
```

src/ansys/geometry/core/tools/problem_areas.py
```diff
@@ -173,12 +184,15 @@ class InexactEdgeProblemAreas(ProblemArea):
         Active supporting geometry service instance for design modeling.
     edges : List[str]
         List of edges associated with the design.
+    modeler : Modeler
+        modeler object
```
RobPasMue, line 188:
```suggestion
        Modeler object.
```

src/ansys/geometry/core/tools/problem_areas.py
```diff
@@ -81,12 +84,15 @@ class DuplicateFaceProblemAreas(ProblemArea):
         Active supporting geometry service instance for design modeling.
     faces : List[str]
         List of faces associated with the design.
+    modeler : Modeler
+        modeler object
```
RobPasMue, line 88:
```suggestion
        Modeler object.
```

src/ansys/geometry/core/tools/problem_areas.py
```diff
@@ -114,6 +120,7 @@ def fix(self) -> RepairToolMessage:
             response.result.created_bodies_monikers,
             response.result.modified_bodies_monikers,
         )
+        self._modeler.read_existing_design()
```
RobPasMue, line 123: Add comments
```suggestion

        # After repair operation we need to sync the active design on the server
        self._modeler.read_existing_design()
```

src/ansys/geometry/core/tools/problem_areas.py
```diff
@@ -129,12 +136,15 @@ class MissingFaceProblemAreas(ProblemArea):
         Active supporting geometry service instance for design modeling.
     edges : List[str]
         List of edges associated with the design.
+    modeler : Modeler
+        modeler object
```
RobPasMue, line 140:
```suggestion
        Modeler object.
```

src/ansys/geometry/core/tools/problem_areas.py
```diff
@@ -202,6 +216,8 @@ def fix(self) -> RepairToolMessage:
             response.result.created_bodies_monikers,
             response.result.modified_bodies_monikers,
         )
+
+        self._modeler.read_existing_design()
```
RobPasMue, line 220:
```suggestion

        # After repair operation we need to sync the active design on the server
        self._modeler.read_existing_design()
```

src/ansys/geometry/core/tools/problem_areas.py
```diff
@@ -271,6 +290,7 @@ def fix(self) -> RepairToolMessage:
             response.result.created_bodies_monikers,
             response.result.modified_bodies_monikers,
         )
+        self._modeler.read_existing_design()
```
RobPasMue, line 293:
```suggestion

        # After repair operation we need to sync the active design on the server
        self._modeler.read_existing_design()
```

src/ansys/geometry/core/tools/problem_areas.py
```diff
@@ -282,16 +302,19 @@ class SplitEdgeProblemAreas(ProblemArea):
     ----------
     id : str
         Server-defined ID for the body.
-    grpc_client : GrpcClient
-        Active supporting geometry service instance for design modeling.
     edges : List[str]
         List of edges associated with the design.
+    grpc_client : GrpcClient
+        Active supporting geometry service instance for design modeling.
```
RobPasMue, line 308:
```suggestion
        Active supporting Geometry Service instance for design modeling.
```

Geometry Service should be capitalized. Please correct everywhere

src/ansys/geometry/core/tools/problem_areas.py
```diff
@@ -315,6 +338,7 @@ def fix(self) -> RepairToolMessage:
             response.result.created_bodies_monikers,
             response.result.modified_bodies_monikers,
         )
+        self._modeler.read_existing_design()
```
RobPasMue, line 341:
```suggestion

        # After repair operation we need to sync the active design on the server
        self._modeler.read_existing_design()
```

src/ansys/geometry/core/tools/problem_areas.py
```diff
@@ -330,12 +354,21 @@ class StitchFaceProblemAreas(ProblemArea):
         Active supporting geometry service instance for design modeling.
     faces : List[str]
         List of faces associated with the design.
+    modeler : Modeler
+        modeler object
```
RobPasMue, line 358:
```suggestion
        Modeler object.
```

src/ansys/geometry/core/tools/problem_areas.py
```diff
@@ -359,4 +392,5 @@ def fix(self) -> RepairToolMessage:
             response.result.created_bodies_monikers,
             response.result.modified_bodies_monikers,
         )
+        self._modeler.read_existing_design()
```
RobPasMue, line 395:
```suggestion

        # After repair operation we need to sync the active design on the server
        self._modeler.read_existing_design()
```

RobPasMue (changes requested):
I would also modify the ``fix()`` functions to accept an optional boolean called ``return_design``, that returns the active design for you, so that you do not have to add an extra line of code as a user in case you want it. By default, it should be false. In case it is false, return ``None``.

Please accept the suggestions provided as well and modify wherever I might have missed something based on my comments. Thanks @umutsoysalansys, great work!

umutsoysalansys:
@RobPasMue fix method is already returning an object, return message. Wouldn't it be too confusing with returning and optional design there?

src/ansys/geometry/core/tools/repair_tools.py
umutsoysalansys: done

RobPasMue:
I wouldn't say so. Python methods can return multiple arguments. My suggestion:

If design is not requested: ``return (message, None)``
If design requested: ``return (message, design)``.

RobPasMue:
Also, there are some broken tests. Please look into them.

smereu:
@RobPasMue @umutsoysalansys @b-matteo I saw the comments about returning a design for a fix operations. I think that it may be useful to have to summarize among us if/how we want to use multiple design in the pyAnsys geometry layer. If I am not mistaken, they mirror the multiple documents in SpaceClaim. As you may know, in Discovery the concept of multiple documents has disappeared (it may come back but there is no immediate plan for it).  Also in SpaceClaim there in assumption that the tools always work on the active document (which in SpaceClaim corresponds to the active window). We seem to have an active document concept here but I would say that a tool should be given a document too (without relying on the active document). I had briefly brought it up with Umut but again, it may be good to be all on the same page on how to deal with it.  I hope that this makes sense!

RobPasMue:
Hi @smereu - multiple documents are handled according to the backend from the Python client. We verify that you are connecting to either a SpaceClaim or Geometry Service backend before allowing the switch to multiple documents. We can discuss this if you prefer nonetheless.

RobPasMue:
@umutsoysalansys - there are tests failing. Please fix them whenever you have the chance

umutsoysalansys:
@RobPasMue @b-matteo Closing this one since we had another change merged to tackle this issue. #https://github.com/ansys/pyansys-geometry/pull/975

PR #936, opened 2024-01-19
fix: verify port availability before attempting connections

### LINKED ISSUE #935: Launching Ansys Desktop products sometimes freezes
When launching the Ansys products in their desktop version, our launcher methods sometimes fail. We have to wait for the port to be available for connection before trying to connect.

RobPasMue:
Closes #935

### DIFF
--- a/src/ansys/geometry/core/connection/product_instance.py
+++ b/src/ansys/geometry/core/connection/product_instance.py
@@ -284,6 +284,11 @@ def prepare_and_start_backend(
 
     instance = ProductInstance(_start_program(args, env_copy).pid)
 
+    # Verify that the backend is ready to accept connections
+    # before returning the Modeler instance.
+    LOG.info("Waiting for backend to be ready...")
+    _wait_for_backend(host, port, timeout)
+
     return Modeler(
         host=host, port=port, timeout=timeout, product_instance=instance, backend_type=backend_type
     )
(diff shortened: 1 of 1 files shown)


src/ansys/geometry/core/connection/product_instance.py
```diff
@@ -298,6 +303,22 @@ def get_available_port():
     return port
 
 
+def _wait_for_backend(host: str, port: int, timeout: int):
+    """Check if the backend is ready to accept connections."""
```
AlejandroFernandezLuces, line 307: Parameters descriptions are missing here

src/ansys/geometry/core/connection/product_instance.py
```diff
@@ -298,6 +303,22 @@ def get_available_port():
     return port
 
 
+def _wait_for_backend(host: str, port: int, timeout: int):
```
AlejandroFernandezLuces, line 306: This is though for local installations right? Maybe we could default the host to `localhost` as in other functions of this module. Also, I think having a default timeout would be nice.

src/ansys/geometry/core/connection/product_instance.py
```diff
@@ -298,6 +303,22 @@ def get_available_port():
...
+    """Check if the backend is ready to accept connections."""
+    import time
+
+    start_time = time.time()
+    while time.time() - start_time < timeout:
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+            if s.connect_ex((host, port)) == 0:
+                return
```
AlejandroFernandezLuces, line 314: A connection message would be nice here
```suggestion
        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
            if s.connect_ex((host, port)) == 0:
                LOG.info("Successful connection to backend.")
                return
```

AlejandroFernandezLuces (approved):
LGTM, some comments

src/ansys/geometry/core/connection/product_instance.py
RobPasMue: Its a private function, default parameters are coming from upstream

src/ansys/geometry/core/connection/product_instance.py
RobPasMue: Agreed, let me add those

src/ansys/geometry/core/connection/product_instance.py
RobPasMue: Yep agreed - but let me modify the message. Connection is not established. We are just verifying that the port is occupied by the service

AlejandroFernandezLuces (approved):

PR #943, opened 2024-01-22
feat: provide script to build your own Geometry Service Docker image

### LINKED ISSUE #915: Instructions on how to build Geometry Service Windows Docker image from local installation
Following https://github.com/ansys/pyansys-geometry/discussions/883

We should provide instructions to users on how to build their local Ansys Geometry Service windows docker image based on a local installation. Pinging @b-matteo for instructions on it. We can discuss offline if needed.

RobPasMue:
As title says. Closes #915

### DIFF
--- a/docker/build_docker_windows.py
+++ b/docker/build_docker_windows.py
@@ -1,0 +1,160 @@
+# Copyright (C) 2023 - 2024 ANSYS, Inc. and/or its affiliates.
+# SPDX-License-Identifier: MIT
+#
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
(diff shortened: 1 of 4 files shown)


RobPasMue:
Requires documenting

docker/build_docker_windows.py
```diff
@@ -0,0 +1,160 @@
...
+            # This version is too old, so we will ignore it
+            continue
+        else:
+            # This version is new enough, so we will add it to the list
+            awp_root[version] = env_val
+
+if len(awp_root) == 0:
+    # There are no Ansys installations
+    print("XXXXXXX No Ansys installations found.. exiting process. XXXXXXX")
+    exit(0)
```
AlejandroFernandezLuces, line 49: If it won't work because the version they have is too old, you should let the user know.

docker/build_docker_windows.py
```diff
@@ -0,0 +1,160 @@
...
+if selection == "":
+    print("... No selection made, using default")
+    selection = len(awp_root)
+else:
+    selection = int(selection)
+print(f">>> Using {list(awp_root.keys())[selection-1]}")
+
+# Get the path to the Ansys installation
+ANSYS_VER = list(awp_root.keys())[selection - 1]
+ANSYS_PATH = list(awp_root.values())[selection - 1]
```
AlejandroFernandezLuces, line 67:
```suggestion
if selection == "":
    print("... No selection made, using default")
    selection = len(awp_root)
else:
    selection = int(selection)
ANSYS_VER = list(awp_root.keys())[selection-1]
print(f">>> Using {ANSYS_VER}")

# Get the path to the Ansys installation
ANSYS_PATH = awp_root[ANSYS_VER]
```

docker/build_docker_windows.py
```diff
@@ -0,0 +1,160 @@
...
+
+# Verify that the Geometry Service is installed
+if not os.path.exists(os.path.join(ANSYS_PATH, "GeometryService")):
+    print("XXXXXXX Geometry Service not installed.. exiting process. XXXXXXX")
+    exit(0)
+
+# Create a temporary directory to copy the Geometry Service files to
+print(">>> Creating temporary directory for building docker image")
+TMP_DIR = os.path.join(os.getcwd(), "docker_geometry_service")
```
AlejandroFernandezLuces, line 76: I think it would be better to use the OS temporary folder for this rather that CWD, for example, `C:\Windows\Temp` and `/temp` in linux, but up to you.

docker/build_docker_windows.py
```diff
@@ -0,0 +1,160 @@
...
+if line != -1:
+    # Get the environment variable name
+    env_var = dockerfile[line : LENGTH_NO_VER + line] + ANSYS_VER
+    # Replace the environment variable with the correct value
+    dockerfile = dockerfile.replace(
+        dockerfile[line : LENGTH_VER + line],
+        env_var,
+    )
+else:
+    print("XXXXXXX No AWP_ROOT environment variable found.. exiting process. XXXXXXX")
```
AlejandroFernandezLuces, line 134:
```suggestion
    print("XXXXXXX No AWP_ROOT environment variable found in Dockerfile.. exiting process. XXXXXXX")
```

AlejandroFernandezLuces (changes requested):

PR #971, opened 2024-02-02
feat: seo-improvements

jorgepiloto:
As per a private discussion, this is setup to avoid creating the default sitemap.xml file.


RobPasMue (approved):
LGTM!

PR #975, opened 2024-02-05
Post repair design sync

umutsoysalansys:
@RobPasMue Another approach on repair sync. Just putting for DuplicateFaces method for simlicity, will apply to the rest. 
Apparently design object has an update itself on it. so no need to bring the modeler into the picture. I had to do a workaround to be able to call the method see desing.py. This looks like a clean update and user don't have to call any additional method after design fix as you can see from the caps below. @b-matteo. Also hidden the exposed face id strings. 

image


src/ansys/geometry/core/tools/problem_areas.py
```diff
@@ -109,11 +112,14 @@ def fix(self) -> RepairToolMessage:
         response = self._repair_stub.FixDuplicateFaces(
             FixDuplicateFacesRequest(duplicate_face_problem_area_id=self._id_grpc)
         )
+        parentdesign = self.faces[0].body.parent_component
```
b-matteo, line 115: @umutsoysalansys this is the way to go but it's not enough yet, since there can be a hierarchy of component, you should iterate on parent_component's parent_component until you get "None" as parent_component's value. When you do, that's your design. If this explanation seems confusing, feel free to reach out, we'll se that together.

RobPasMue:
> uplicateFaces method for simlicity, will apply to the rest. Apparently design object has an update itself on it. so no need to bring the modeler into the picture. I had to do a workaround to be able to call the method see desing.py. This looks like a clean update and user don't have to call any additional method after design fix as you can see from the caps below. @b-matteo. Also hidden the exposed face id strings.

It's looking towards the right approach, but as @b-matteo mentioned - you need to handle properly the tree, abstraction, etc. Let me know if you need anything.

RobPasMue:
I'll start reviewing this morning..

src/ansys/geometry/core/tools/repair_tools.py
```diff
@@ -254,8 +289,106 @@ def find_stitch_faces(self, bodies: List["Body"]) -> List[StitchFaceProblemAreas
...
+        -------
+        List[Edge]
+            List of edge objects representing duplicate face problem areas.
+        """
+        return [
+            edge
+            for body in design.bodies
+            for face in body.faces
+            for edge in face.edges
+            if str(edge.id) in set(map(str, edge_ids))
```
RobPasMue, line 373: Why are you using set(map(...)) instead of using the list of ``edge_ids`` directly to verify whether the id is inside or not? Also, why are you doing str(edge.id)? It is already a string...

RobPasMue (changes requested):
Needs several big changes @umutsoysalansys - especially the way you are reading the existing design.. if you want to completely read a new design from the service, you should always use the ``Modeler``.

RobPasMue:
Have a look at all my changes @umutsoysalansys - this should be a more consistent and modular approach to your implementation for updating inplace the design. I also extracted common functionalities and implemented sanity checks on inputs. This should serve your purposes now. @b-matteo @AlejandroFernandezLuces please review since I might have done something wrong.

src/ansys/geometry/core/designer/design.py
```diff
@@ -759,3 +762,35 @@ def __read_existing_design(self) -> None:
...
+        #       the two methods are in sync.
+        #
+        self._components = []
+        self._bodies = []
+        self._materials = []
+        self._named_selections = {}
+        self._coordinate_systems = {}
+
+        # Get the previous design id
+        previous_design_id = self._design_id
```
umutsoysalansys, line 788: @RobPasMue , so why this id check needed? @b-matteo

umutsoysalansys:
@RobPasMue, so no need to use modeler here as you said above? I thought this was the way you already agreed upon. Thanks for refactoring, looks good to me. @b-matteo

RobPasMue:
> @RobPasMue, so no need to use modeler here as you said above? I thought this was the way you already agreed upon. Thanks for refactoring, looks good to me. @b-matteo

No need. As I mentioned to you offline, I got it the other way around. My bad.

src/ansys/geometry/core/designer/design.py
RobPasMue: This will be needed for future cases... Not only repair tools in which the ID is kept the same.

RobPasMue:
Repair tools require examples and documentation indeed - @umutsoysalansys please consider doing that shortly. Otherwise all these tools are useless for end users since they won't know how to use them

RobPasMue (approved):

RobPasMue:
@AlejandroFernandezLuces - please review

PR #1018, opened 2024-02-26
fix: canonical link in pages

jorgepiloto:
Includes a canonical link in every HTML page except `announcement.html` and `webpack-macros.html` since these are template files.


RobPasMue (approved):
LGTM!

Issue #1004, opened 2024-02-21
Implement the usage of ``pyansys-tools-versioning`` or own decorator for minimum server version


RobPasMue:
As title says.

RobPasMue:
This is now possible using ``ansys-api-geometry`` version 0.3.8

RobPasMue:
We will have to implement some try/except behavior though to avoid failing past versions. But in those cases, we will assume 24R1.

RobPasMue:
Done through #1024

PR #1024, opened 2024-02-28
feat: Add min_backend_version decorator

AlejandroFernandezLuces:
## Description
Add a minimum version decorator to allow checking if methods are compatible with the current backend.

## Issue linked
#1004

## Checklist

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -12,4 +12,5 @@
 - [ ] I have reviewed my changes before submitting this pull request.
 - [ ] I have linked the issue or issues that are solved to the PR if any.
 - [ ] I have assigned this PR to myself.
+- [ ] I have added the minimum version decorator to any new backend method implemented.
 - [ ] I have made sure that the title of my PR follows [Conventional commits style](https://www.conventionalcommits.org/en/v1.0.0/#summary) (e.g. ``feat: extrude circle to cylinder``)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ dependencies = [
     "scipy>=1.7.3,<2",
     "six>=1.16.0,<2",
     "vtk>=9,<10",
-    "semver>=3.0.2",
+    "semver>=3,<4",
 ]
 
 [project.optional-dependencies]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,6 +66,7 @@ tests = [
     "pyvista[jupyter]==0.43.3",
     "requests==2.31.0",
     "scipy==1.12.0",
+    "semver==3.0.2",
     "six==1.16.0",
     "vtk==9.3.0",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,6 +38,7 @@ dependencies = [
     "scipy>=1.7.3,<2",
     "six>=1.16.0,<2",
     "vtk>=9,<10",
+    "semver>=3.0.2",
 ]
 
 [project.optional-dependencies]
--- a/src/ansys/geometry/core/connection/client.py
+++ b/src/ansys/geometry/core/connection/client.py
@@ -172,7 +172,7 @@ class GrpcClient:
 
         # retrieve the backend information
         grpc_backend_response = self._admin_stub.GetBackend(Empty())
-        
+
         # if no backend type has been specified, ask the backend which type it is
         if backend_type == None:
             grpc_backend_type = grpc_backend_response.type
--- a/src/ansys/geometry/core/connection/client.py
+++ b/src/ansys/geometry/core/connection/client.py
@@ -27,6 +27,7 @@ import time
 
 from ansys.api.dbu.v0.admin_pb2 import BackendType as GRPCBackendType
 from ansys.api.dbu.v0.admin_pb2_grpc import AdminStub
+from ansys.api.dbu.v0.dbumodels_pb2 import VersionIdentifier
 from beartype import beartype as check_input_types
 from beartype.typing import Optional, Union
 from google.protobuf.empty_pb2 import Empty
@@ -170,8 +171,9 @@ class GrpcClient:
         self._admin_stub = AdminStub(self._channel)
 
         # if no backend type has been specified, ask the backend which type it is
+        grpc_backend_response = self._admin_stub.GetBackend(Empty())
         if backend_type == None:
-            grpc_backend_type = self._admin_stub.GetBackend(Empty()).type
+            grpc_backend_type = grpc_backend_response.type
             if grpc_backend_type == GRPCBackendType.DISCOVERY:
                 backend_type = BackendType.DISCOVERY
             elif grpc_backend_type == GRPCBackendType.SPACECLAIM:
@@ -181,6 +183,13 @@ class GrpcClient:
             elif grpc_backend_type == GRPCBackendType.LINUX_DMS:
                 backend_type = BackendType.LINUX_SERVICE
 
+        # backend version
+        if hasattr(grpc_backend_response, "version"):
+            ver = grpc_backend_response.version
+            self._backend_version = f"{ver.major_release}.{ver.minor_release}.{ver.service_pack}"
+        else:
+            logger.warning("The backend version is only available after 24.1 version.")
+            self._backend_version = "24.1.0"
         # Store the backend type
         self._backend_type = backend_type
         self._multiple_designs_allowed = (
@@ -201,6 +210,18 @@ class GrpcClient:
         not straightforward.
         """
         return self._backend_type
+
+    @property
+    def backend_version(self) -> VersionIdentifier:
+        """
+        Backend version.
+
+        Notes
+        -----
+        This method might return ``None`` because determining the backend version is
+        not straightforward.
+        """
+        return self._backend_version
 
     @property
     def multiple_designs_allowed(self) -> bool:
--- a/src/ansys/geometry/core/connection/client.py
+++ b/src/ansys/geometry/core/connection/client.py
@@ -27,7 +27,6 @@ import time
 
 from ansys.api.dbu.v0.admin_pb2 import BackendType as GRPCBackendType
 from ansys.api.dbu.v0.admin_pb2_grpc import AdminStub
-from ansys.api.dbu.v0.dbumodels_pb2 import VersionIdentifier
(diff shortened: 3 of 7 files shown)


src/ansys/geometry/core/misc/checks.py
```diff
@@ -302,3 +303,44 @@ def check_type_all_elements_in_iterable(
     """
     for elem in input:
         check_type(elem, expected_type)
+
+
+def min_backend_version(method_version: str):
+    """
+    Compare the min version required by a method to the current backend version.
```
PipKat, line 310:
```suggestion
    Compare a method's minimum required version to the current backend version.
```

src/ansys/geometry/core/misc/checks.py
```diff
@@ -302,3 +303,44 @@ def check_type_all_elements_in_iterable(
...
+
+    Parameters
+    ----------
+    method_version : str
+        The minimum version required by the method.
```
PipKat, line 315:
```suggestion
        Minimum version required by the method.
```

src/ansys/geometry/core/misc/checks.py
```diff
@@ -302,3 +303,44 @@ def check_type_all_elements_in_iterable(
...
+        If the client is not available.
+    """
+    # Lazy import to avoid circular imports
+    from ansys.geometry.core.errors import GeometryRuntimeError
+
+    def backend_version_decorator(method):
+        def wrapper(self, *args, **kwargs):
+            if self._grpc_client is None:
+                raise GeometryRuntimeError(
+                    "The client is not available. Please initialize the client first."
```
PipKat, line 331:
```suggestion
                    "The client is not available. You must initialize the client first."
```

.github/pull_request_template.md
```diff
@@ -12,4 +12,5 @@
 - [ ] I have reviewed my changes before submitting this pull request.
 - [ ] I have linked the issue or issues that are solved to the PR if any.
 - [ ] I have assigned this PR to myself.
+- [ ] I have added the minimum version decorator to any new backend method implemented.
```
RobPasMue, line 15: Great idea - thanks @AlejandroFernandezLuces

pyproject.toml
```diff
@@ -38,6 +38,7 @@ dependencies = [
     "scipy>=1.7.3,<2",
     "six>=1.16.0,<2",
     "vtk>=9,<10",
+    "semver>=3.0.2",
```
RobPasMue, line 41:
```suggestion
    "semver>=3,<4",
```
We should block major versions on general deps

RobPasMue: We should also add a pinned version to the "tests" extras

src/ansys/geometry/core/connection/client.py
```diff
@@ -170,8 +171,9 @@ def __init__(
         self._admin_stub = AdminStub(self._channel)
 
         # if no backend type has been specified, ask the backend which type it is
+        grpc_backend_response = self._admin_stub.GetBackend(Empty())
         if backend_type == None:
```
RobPasMue, line 175:
```suggestion
        # retrieve the backend information
        grpc_backend_response = self._admin_stub.GetBackend(Empty())
        
        # if no backend type has been specified, ask the backend which type it is
        if backend_type == None:
```

src/ansys/geometry/core/connection/client.py
```diff
@@ -181,6 +183,13 @@ def __init__(
             elif grpc_backend_type == GRPCBackendType.LINUX_DMS:
                 backend_type = BackendType.LINUX_SERVICE
 
+        # backend version
```
RobPasMue, line 186:
```suggestion
        # retrieve the backend version
```

src/ansys/geometry/core/connection/client.py
```diff
@@ -181,6 +183,13 @@ def __init__(
...
+        if hasattr(grpc_backend_response, "version"):
+            ver = grpc_backend_response.version
+            self._backend_version = f"{ver.major_release}.{ver.minor_release}.{ver.service_pack}"
+        else:
+            logger.warning("The backend version is only available after 24.1 version.")
+            self._backend_version = "24.1.0"
         # Store the backend type
```
RobPasMue, line 193:
```suggestion
            self._backend_version = "24.1.0"

        # Store the backend type
```

src/ansys/geometry/core/connection/client.py
```diff
@@ -202,6 +211,18 @@ def backend_type(self) -> BackendType:
         """
         return self._backend_type
 
+    @property
+    def backend_version(self) -> VersionIdentifier:
```
RobPasMue, line 215:
```suggestion
    def backend_version(self) -> str:
```

src/ansys/geometry/core/connection/client.py
```diff
@@ -202,6 +211,18 @@ def backend_type(self) -> BackendType:
...
+        """
+        Get the current backend version.
+
+        Returns
+        -------
+        VersionIdentifier
```
RobPasMue, line 221:
```suggestion
        str
```

src/ansys/geometry/core/connection/client.py
```diff
@@ -202,6 +211,18 @@ def backend_type(self) -> BackendType:
...
+            Backend version.
```
RobPasMue, line 222:
```suggestion
            Backend version in semantic versioning format (that is, Ansys 24R1 SP2 would be ``24.1.2``).
```

src/ansys/geometry/core/misc/checks.py
```diff
@@ -302,3 +303,44 @@ def check_type_all_elements_in_iterable(
...
+    GeometryRuntimeError
+        If the method version is higher than the backend version.
+    GeometryRuntimeError
+        If the client is not available.
+    """
+    # Lazy import to avoid circular imports
+    from ansys.geometry.core.errors import GeometryRuntimeError
+
+    def backend_version_decorator(method):
+        def wrapper(self, *args, **kwargs):
```
RobPasMue, line 328: Let's verify that "self" has an attribute called "_grpc_client". It should always be the case but it will also act as a "defensive" measure for developers.

src/ansys/geometry/core/misc/checks.py
```diff
@@ -302,3 +303,44 @@ def check_type_all_elements_in_iterable(
...
+            if self._grpc_client is None:
+                raise GeometryRuntimeError(
```
RobPasMue, line 330: I'm unsure whether there is any scenario where the client is not defined (as long as the attribute exists). We always pass it in construction to all objects. But in any case, good "defensive" measure

src/ansys/geometry/core/connection/client.py
```diff
@@ -181,6 +183,13 @@ def __init__(
...
 
+        # backend version
+        if hasattr(grpc_backend_response, "version"):
+            ver = grpc_backend_response.version
+            self._backend_version = f"{ver.major_release}.{ver.minor_release}.{ver.service_pack}"
+        else:
+            logger.warning("The backend version is only available after 24.1 version.")
+            self._backend_version = "24.1.0"
         # Store the backend type
         self._backend_type = backend_type
```
RobPasMue, line 194: I would reorder this section so that the "backend_type" logic is all together, and the backend version logic comes right afterwards

RobPasMue (changes requested):
Overall looking good - some minor changes. Request review again once addressed. Great work @AlejandroFernandezLuces

RobPasMue (commented):
I am missing some tests by the way. Can you test the decorator on its own? Use monkeypatch to avoid having to adapt the test in the future. Create a random object that has a "_grpc_client" attribute and it also has another attribute called backend version. Testing is required for the decorator only

AlejandroFernandezLuces:
Added the test, and fixed some issues that arose from it. Also, it seems that is only working on the unstable version, I guess that is due to the feature not being available in the stable version?

RobPasMue:
> Added the test, and fixed some issues that arose from it. Also, it seems that is only working on the unstable version, I guess that is due to the feature not being available in the stable version?

Yep - precisely. This will require a new stable release. Let me trigger it

RobPasMue:
Rerunning all jobs - images have been published

AlejandroFernandezLuces:
Thanks for the reviews @PipKat @RobPasMue 😄 

Enabling automerge.

pyproject.toml
```diff
@@ -38,6 +38,7 @@ dependencies = [
     "scipy>=1.7.3,<2",
     "six>=1.16.0,<2",
     "vtk>=9,<10",
+    "semver>=3,<4",
```
RobPasMue, line 41: Add a pinned version in the tests dependencies

src/ansys/geometry/core/misc/checks.py
```diff
@@ -302,3 +303,48 @@ def check_type_all_elements_in_iterable(
     """
     for elem in input:
         check_type(elem, expected_type)
+
+
+def min_backend_version(method_version: str):
```
RobPasMue, line 308: I'm thinking... why not use 3 integers to define the backend version? major, minor an sp version, And we build the string on our own inside of it.

I'm just thinking of a wrongly-formatted decorator - string input is very flexible. However integers can only be integers. Anyway, what do you think?

RobPasMue (changes requested):
Tests look really good. One suggestion and two change requests. After that we can merge

src/ansys/geometry/core/misc/checks.py
AlejandroFernandezLuces: Agree, I prefer your suggestion.

RobPasMue (commented):
This solves the test issues

tests/test_misc_checks.py
```diff
@@ -199,3 +201,54 @@ def test_check_ndarray_is_not_none():
...
+    with pytest.raises(
+        GeometryRuntimeError,
+        match="The method 'case_higher' is not available for the current backend version 24.1.0.",
+    ):
+        case_higher(mock_object)
+
+    # If client is not initialized, an error should be raised
+    mock_object._grpc_client = None
+
+    @min_backend_version("24.1.0")
```
RobPasMue, line 246:
```suggestion
    @min_backend_version(24, 1, 0)
```

RobPasMue (approved):

RobPasMue (approved):

Issue #1036, opened 2024-03-05
Scale and Mirror a body


jonahrb:
### 📝 Description of the feature

similar to how we have `body.rotate()` and `body.translate()` we should add `body.scale()` and `body.mirror()`

### 💡 Steps for implementing the feature

`body.scale()` should just take a float and will use `API.Geometry.Matrix.CreateScale()` on ApiServer

I'm assuming `body.mirror()` should take a plane and use `API.Geometry.Matrix.CreateMapping()`

jonahrb:
I considered this a while ago - the ability to allow the user to supply their own 4x4 transformation matrix, instead of performing each operation individually. That can be discussed again during the implementation of this feature request.

RobPasMue:
Hi @jonahrb   let's implement these methods! About the 4x4 by matrix, we can give it another thought. I believe that with this full set of methods, everything is available. Nonetheless, if they want to cut their way and do a full transformation that would be possible... but it might be error prone. We can discuss offline

PR #1045, opened 2024-03-07
feat: scale and map body

jonahrb:
## Description
Ability to scale or map a body

## Issue linked
#1036 

## Checklist


src/ansys/geometry/core/designer/body.py
```diff
@@ -358,6 +362,42 @@ def rotate(
         """
         return
 
+    @abstractmethod
+    def scale(
+        self,
+        value: Real,
+    ) -> None:
```
RobPasMue, line 369: I'd single line them if possible
```suggestion
    def scale(self, value: Real) -> None:
```

src/ansys/geometry/core/designer/body.py
```diff
@@ -358,6 +362,42 @@ def rotate(
...
+        """
+        Scale the geometry body by the given value.
+
+        Parameters
+        ----------
+        value: Real
+            Value to scale the body by.
+        Returns
+        -------
+        None
```
RobPasMue, line 379: If nothing is returned, no need to add the return section

```suggestion
            Value to scale the body by.
```

src/ansys/geometry/core/designer/body.py
```diff
@@ -358,6 +362,42 @@ def rotate(
...
+        """
+        return
+
+    @abstractmethod
+    def mirror(
+        self,
+        frame: Frame,
+    ) -> None:
```
RobPasMue, line 387:
```suggestion
    @abstractmethod
    def mirror(self, frame: Frame) -> None:
```

src/ansys/geometry/core/designer/body.py
```diff
@@ -358,6 +362,42 @@ def rotate(
...
+        """
+        Mirror the geometry body across the specified frame.
+
+        Parameters
+        ----------
+        frame: Frame
+            Structure defining the mirroring/mapping of the body.
+        Returns
+        -------
+        None
```
RobPasMue, line 397:
```suggestion
            Structure defining the mirroring/mapping of the body.
```

src/ansys/geometry/core/designer/body.py
```diff
@@ -818,6 +858,25 @@ def rotate(
             )
         )
 
+    @protect_grpc
+    @check_input_types
+    @reset_tessellation_cache
+    @min_backend_version(24, 1, 0)
```
RobPasMue, line 864: Oh! It was already available for all backends?

src/ansys/geometry/core/designer/body.py
```diff
@@ -818,6 +858,25 @@ def rotate(
...
+        self._bodies_stub.Scale(ScaleRequest(id=self.id, scale=value))
+
+    @protect_grpc
+    @check_input_types
+    @reset_tessellation_cache
+    @min_backend_version(24, 1, 0)
+    def mirror(
+        self,
+        frame: Frame,
+    ) -> None:  # noqa: D102
```
RobPasMue, line 876:
```suggestion
    def mirror(self, frame: Frame) -> None:  # noqa: D102
```

src/ansys/geometry/core/designer/body.py
RobPasMue, line 864: If that is the case (it was already available in 24R1) no need to add the decorator. We assume the first version is 24R1.
```suggestion
```

src/ansys/geometry/core/designer/body.py
```diff
@@ -818,6 +858,25 @@ def rotate(
...
+    @reset_tessellation_cache
+    @min_backend_version(24, 1, 0)
+    def scale(self, value: Real) -> None:  # noqa: D102
+        self._grpc_client.log.debug(f"Scaling body {self.id}.")
+        self._bodies_stub.Scale(ScaleRequest(id=self.id, scale=value))
+
+    @protect_grpc
+    @check_input_types
+    @reset_tessellation_cache
+    @min_backend_version(24, 1, 0)
```
RobPasMue, line 872: Same as before here
```suggestion
```

RobPasMue (approved):
Looking good @jonahrb

src/ansys/geometry/core/designer/body.py
jonahrb: @RobPasMue No, but i was able to retroactively add it, so it is technically available with 24.1 but they would need an updated service :)

RobPasMue: In that case... let's avoid "retroactive" behavior. I talked about this with @b-matteo. Your service is the one that defines the backend version number so you need at least a 24R2 service to do this. Meaning that we need to add the decorators as 24, 2, 0

jonahrb: @RobPasMue makes sense, on it

jonahrb:
@RobPasMue FYI i renamed "mirror" to "map" since that's a more accurate description of the method. I also added unit tests, see `test_body_mapping()` for a couple examples

tests/integration/test_design.py
```diff
@@ -1927,3 +1927,111 @@ def test_get_collision(modeler: Modeler):
 
     assert body1.get_collision(body2) == CollisionType.TOUCH
     assert body2.get_collision(body3) == CollisionType.NONE
+
+
+def test_body_scale(modeler: Modeler):
+    """Test for verifying the correct scaling of a ``Body``."""
```
PipKat, line 1933:
```suggestion
    """Verify the correct scaling of a body."""
```

tests/integration/test_design.py
```diff
@@ -1927,3 +1927,111 @@ def test_get_collision(modeler: Modeler):
...
+
+    body.scale(2)
+    assert Accuracy.length_is_equal(body.volume.m, 8)
+
+    body.scale(0.25)
+    assert Accuracy.length_is_equal(body.volume.m, 1 / 8)
+
+
+def test_body_mapping(modeler: Modeler):
+    """Test for verifying the correct mapping of a ``Body``."""
```
PipKat, line 1948:
```suggestion
    """Verify the correct mapping of a body."""
```

tests/integration/test_design.py
RobPasMue: a body works @PipKat - thanks for pointing it out nonetheless :)

RobPasMue (dismissed):
Tests are great, implementation is great. Thanks a lot @jonahrb - this is going to be extremely helpful. I would only suggest adding some examples in our documentation on how to do these operations. But it can be done in another PR. I will create an issue and assign it to you.

dastan-ansys (dismissed):
looks good! I am not big on documentation personally, but I feel like it would be a good idea to add a comment that describes the concrete implementation of scale() and map() or links them back the comment on their abstract counterparts.

b-matteo (approved):

PR #1171, opened 2024-04-26
fix: GetSurface and GetCurve not available prior to 24R2

### LINKED ISSUE #1170: GetSurface is not implemented in 24R1 causing latest release of PyAnsys Geometry to fail
We cannot request face.normal() anymore in 24R1 and using the latest PyAnsys Geometry version - the gRPC API GetSurface is not implemented until 24R2. We need to fallback to the old way in case the backend is older than 24R2

RobPasMue:
## Description
Fixing 24R1 related issue due to GetSurface and GetCurve APIs

## Issue linked
Closes #1170 

## Checklist

### DIFF
--- a/src/ansys/geometry/core/designer/edge.py
+++ b/src/ansys/geometry/core/designer/edge.py
@@ -107,7 +107,7 @@ class Edge:
         return self._is_reversed
 
     @property
-    @min_backend_version(24,2,0)
+    @min_backend_version(24, 2, 0)
     def shape(self) -> TrimmedCurve:
         """
         Underlying trimmed curve of the edge.
@@ -145,7 +145,7 @@ class Edge:
         """Calculated length of the edge."""
         try:
             return self.shape.length
-        except GeometryRuntimeError: # pragma: no cover
+        except GeometryRuntimeError:  # pragma: no cover
             # Only for versions < 24.2.0 (i.e. before the introduction of the shape property)
             self._grpc_client.log.debug("Requesting edge length from server.")
             length_response = self._edges_stub.GetLength(self._grpc_id)
--- a/src/ansys/geometry/core/designer/edge.py
+++ b/src/ansys/geometry/core/designer/edge.py
@@ -146,7 +146,7 @@ class Edge:
         try:
             return self.shape.length
         except GeometryRuntimeError:  # pragma: no cover
-            # Only for versions < 24.2.0 (i.e. before the introduction of the shape property)
+            # Only for versions earlier than 24.2.0 (before the introduction of the shape property)
             self._grpc_client.log.debug("Requesting edge length from server.")
             length_response = self._edges_stub.GetLength(self._grpc_id)
             return Quantity(length_response.length, DEFAULT_UNITS.SERVER_LENGTH)
--- a/src/ansys/geometry/core/designer/edge.py
+++ b/src/ansys/geometry/core/designer/edge.py
@@ -30,9 +30,9 @@ from pint import Quantity
 
 from ansys.geometry.core.connection.client import GrpcClient
 from ansys.geometry.core.connection.conversions import grpc_curve_to_curve
-from ansys.geometry.core.errors import protect_grpc
+from ansys.geometry.core.errors import GeometryRuntimeError, protect_grpc
 from ansys.geometry.core.math.point import Point3D
-from ansys.geometry.core.misc.checks import ensure_design_is_active
+from ansys.geometry.core.misc.checks import ensure_design_is_active, min_backend_version
 from ansys.geometry.core.misc.measurements import DEFAULT_UNITS
 from ansys.geometry.core.shapes.curves.trimmed_curve import ReversedTrimmedCurve, TrimmedCurve
 from ansys.geometry.core.shapes.parameterization import Interval
@@ -107,6 +107,7 @@ class Edge:
         return self._is_reversed
 
     @property
+    @min_backend_version(24,2,0)
     def shape(self) -> TrimmedCurve:
         """
         Underlying trimmed curve of the edge.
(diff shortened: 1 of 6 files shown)


src/ansys/geometry/core/designer/face.py
```diff
@@ -196,6 +200,7 @@ def body(self) -> "Body":
         return self._body
 
     @property
+    @min_backend_version(24, 2, 0)
```
AlejandroFernandezLuces, line 203: This will raise an error if the version is lesser than 24.2 before launching the function right? I think you might never reach the except of the try in this function because of that

src/ansys/geometry/core/designer/edge.py
```diff
@@ -142,7 +143,13 @@ def shape(self) -> TrimmedCurve:
     @ensure_design_is_active
     def length(self) -> Quantity:
```
AlejandroFernandezLuces, line 144: Is the decorator needed here too?

AlejandroFernandezLuces (dismissed):
LGTM, just those two minor comments

src/ansys/geometry/core/designer/edge.py
RobPasMue: No, there is a try except inside. If it fails due to the backend version error, we fallback to the old implementation.

src/ansys/geometry/core/designer/face.py
RobPasMue: The decorator is for the ``shape`` property. The try/except is on the ``normal`` function

RobPasMue:
I believe the comments don't apply @AlejandroFernandezLuces - please review my answers

src/ansys/geometry/core/designer/face.py
AlejandroFernandezLuces: Misread the diff 😅

b-matteo (approved):
Apart from @AlejandroFernandezLuces 's comment it looks good to me. Thanks @RobPasMue

src/ansys/geometry/core/designer/face.py
RobPasMue: Yeah the diff was a bit ugly - I know 😄

RobPasMue:
Thank you both for your reviews!

src/ansys/geometry/core/misc/checks.py
```diff
@@ -331,19 +331,19 @@ def min_backend_version(major: int, minor: int, service_pack: int):
...
                     )
                 elif self._grpc_client.backend_version is not None:
-                    comp = semver.compare(method_version, self._grpc_client.backend_version)
+                    comp = method_version.compare(self._grpc_client.backend_version)
                     # if comp is 1, method version is higher than backend version.
                     if comp == 1:
 
                         # Check if the version is "0.0.0" (i.e., the version is not available)
-                        if self._grpc_client.backend_version == "0.0.0":
+                        if str(self._grpc_client.backend_version) == "0.0.0":
```
AlejandroFernandezLuces, line 346: Nitpicking, maybe is better to use the semver package directly to compare instead of strings? I couldn't find an exact version comparison, but we can use this. It is a bit more verbose, up to you:
```suggestion
                        if self._grpc_client.backend_version.is_compatible(semver.Version(0, 0, 0)):
```

AlejandroFernandezLuces (approved):
This is better handled with the `semver` package, nice addition 🙂

src/ansys/geometry/core/designer/face.py
```diff
@@ -333,7 +344,13 @@ def point(self, u: float = 0.5, v: float = 0.5) -> Point3D:
             :class:`Point3D`
             object evaluated at the given UV coordinates.
         """
-        return self.shape.evaluate_proportion(u, v).position
+        try:
+            return self.shape.evaluate_proportion(u, v).position
+        except GeometryRuntimeError:  # pragma: no cover
+            # Only for versions < 24.2.0 (i.e. before the introduction of the shape property)
```
PipKat, line 350:
```suggestion
            # Only for versions earlier than 24.2.0 (before the introduction of the shape property)
```

src/ansys/geometry/core/designer/edge.py
```diff
@@ -142,7 +143,13 @@ def shape(self) -> TrimmedCurve:
...
         """Calculated length of the edge."""
-        return self.shape.length
+        try:
+            return self.shape.length
+        except GeometryRuntimeError:  # pragma: no cover
+            # Only for versions < 24.2.0 (i.e. before the introduction of the shape property)
```
PipKat, line 149:
```suggestion
            # Only for versions earlier than 24.2.0 (before the introduction of the shape property)
```

src/ansys/geometry/core/designer/face.py
```diff
@@ -305,7 +310,13 @@ def normal(self, u: float = 0.5, v: float = 0.5) -> UnitVector3D:
             This :class:`UnitVector3D` object is perpendicular to the surface at the
             given UV coordinates.
         """
-        return self.shape.normal(u, v)
+        try:
+            return self.shape.normal(u, v)
+        except GeometryRuntimeError:  # pragma: no cover
+            # Only for versions < 24.2.0 (i.e. before the introduction of the shape property)
```
PipKat, line 316:
```suggestion
            # Only for versions earlier than 24.2.0 (before the introduction of the shape property)
```

src/ansys/geometry/core/misc/checks.py
RobPasMue: Hmm got your point... but I'd prefer to avoid using the Version object there... more time consuming plus it is an edge case honestly (0.0.0 shouldn't happen starting on 24R2... but it might happen A LOT in older versions - i.e. 24R1). The string representation of the object you get it for free - so since it is such an overwhelmingly called method, I'd avoid doing that. Thanks for the suggestion nonetheless!

PR #1264, opened 2024-06-17
fix: unnecessary Point3D comparison

RobPasMue:
Potential bug... array comparison shouldn't be happening. You just want to verify whether the value is 0, to perform a 90º projection.

### DIFF
--- a/src/ansys/geometry/core/shapes/surfaces/sphere.py
+++ b/src/ansys/geometry/core/shapes/surfaces/sphere.py
@@ -197,7 +197,7 @@ class Sphere(Surface):
         x = origin_to_point.dot(self.dir_x)
         y = origin_to_point.dot(self.dir_y)
         z = origin_to_point.dot(self.dir_z)
-        if np.allclose((x * x + y * y + z * z), 0):
+        if np.allclose(x * x + y * y + z * z, 0):
             return SphereEvaluation(self, ParamUV(0, np.pi / 2))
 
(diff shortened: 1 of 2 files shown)


src/ansys/geometry/core/shapes/surfaces/sphere.py
```diff
@@ -197,7 +197,7 @@ def project_point(self, point: Point3D) -> "SphereEvaluation":
         x = origin_to_point.dot(self.dir_x)
         y = origin_to_point.dot(self.dir_y)
         z = origin_to_point.dot(self.dir_z)
-        if np.allclose((x * x + y * y + z * z), Point3D([0, 0, 0])):
+        if np.allclose((x * x + y * y + z * z), 0):
```
AlejandroFernandezLuces, line 200: From numpy documentation:
> Returns True if two arrays are element-wise equal within a tolerance.
Parameters: a, b array_like

It seems that `allclose` uses two arrays. 

I would use the same data type for both values, ideally some form of array.
```suggestion
        if np.allclose(x * x + y * y + z * z, 0):
```
or
```suggestion
        if np.allclose((x * x + y * y + z * z), (0)):
```

RobPasMue: Oops. Totally right, forgot to remove the parenthesis on the square product. Thanks @AlejandroFernandezLuces

PR #1279, opened 2024-06-24
build: update sphinx-autodoc-typehints version

RobPasMue:
As title says


RobPasMue:
@dependabot unignore sphinx-autodoc-typehints

PR #1274, opened 2024-06-20
feat: using ruff as the main linter/formatter

RobPasMue:
## Description
Switching to Ruff

## Issue linked
Closes #1272 

## Checklist
- [X] I have tested my changes locally.
- [X] I have added necessary documentation or updated existing documentation.
- [X] I have followed the coding style guidelines of this project.
- [X] I have added appropriate unit tests.
- [X] I have reviewed my changes before submitting this pull request.
- [X] I have linked the issue or issues that are solved to the PR if any.
- [X] I have assigned this PR to myself.
- [X] I have added the minimum version decorator to any new backend method implemented.
- [X] I have made sure that the title of my PR follows [Conventional commits style](https://www.conventionalcommits.org/en/v1.0.0/#summary) (e.g. ``feat: extrude circle to cylinder``)

PR #1289, opened 2024-07-02
fix: manifest path should render as posix rather than uri

RobPasMue:
As title says - this was causing the impossibility to launch SpaceClaim or Discovery with PyAnsys Geometry. Coming from #1274 ...


RobPasMue:
Pinging #1285 and @XiongHuyu - kudos for bringing the issue to our attention

Issue #433, opened 2023-03-02
PLOTTING: color alignment with service(Discovery,SpaceClaim)


RobPasMue:
As part of the JSON information, whenever it is send back, we also get information on the color of each body/component. We should make use of it when plotting.

RobPasMue:
Partially possible via #1357 - now we have to implement the plotter logic to use this color value.

PR #1357, opened 2024-08-12
feat: get and set body color

umutsoysalansys:
## Description
Summary:
This pull request introduces new functionality to the geometry core, enabling users to get and set the color of a Body object. This enhancement extends the existing API by adding properties and methods that allow for seamless color management within the design environment.

New Features:
Get Color: Users can now retrieve the current color of a Body object using the color property. This property will return the color in a hexadecimal format (e.g., "#0000FF").

Set Color: The set_color method allows users to set the color of a Body object. The method accepts color values in various formats:

Formats supported: 
Hexadecimal color codes (e.g., "#0000FF").
Named color strings (e.g., "blue").
RGB values as a tuple, either in the range (0-1) as floats or (0-255) as integers.

image.

## Usage Example 
```python
# Create the sphere
sphere = design.create_sphere("particle", Point3D([0.0, 0.0, 0.0], unit=unit), Distance(0.5))

# Import matplotlib.colors
import matplotlib.colors as mcolors

# Create a blue color using RGBA and assign it to the sphere
blue_color = mcolors.to_rgba("#0000FF")
sphere.set_color(blue_color)

# Set color directly from a hex code
sphere.set_color("#0000FF")

# Set color directly using RGB integers (values between 0-255)
sphere.set_color((0, 255, 0))

# Set the color using a named color
sphere.set_color("green")

# Set color using RGB floats (values between 0.0-1.0)
sphere.set_color((1.0, 0.0, 0.0))
```

## Issue linked
#433 related but also recent customer feedback. 
## Checklist


RobPasMue:
Hey @umutsoysalansys - you could use matplotlib for this purpose. See this example:

```py
import matplotlib.colors as mcolors

# Convert RGB to HEX
rgb = (0.5, 0.5, 0.5)  # values between 0 and 1
hex_code = mcolors.to_hex(rgb)
print(hex_code)  # Output: '#808080'

# Convert color name to HEX
color_name = 'blue'
hex_code = mcolors.to_hex(mcolors.CSS4_COLORS[color_name])
print(hex_code)  # Output: '#0000FF'

# Convert HEX to HEX (just returns the same HEX code)
hex_code = '#ff6347'
hex_code = mcolors.to_hex(hex_code)
print(hex_code)  # Output: '#ff6347'
```

src/ansys/geometry/core/designer/body.py
```diff
@@ -686,6 +697,7 @@ def __init__(
         self._commands_stub = CommandsStub(self._grpc_client.channel)
         self._tessellation = None
         self._fill_style = FillStyle.DEFAULT
+        self._color = None
```
RobPasMue, line 700: Would it be possible to get the information from the server on which color is being used? Even if this only works on 25R1 and onwards. We could implement it as a lazy loaded property (we assign it to None by default and if a user requests the current color, and it is None... we go to the server to retrieve it). And if it is already defined we just return the value. What's your take @umutsoysalansys? :)

RobPasMue:
Since unstable image is working, I am pushing image to stable

src/ansys/geometry/core/designer/body.py
umutsoysalansys: implemented

tests/integration/test_design.py
```diff
@@ -2010,6 +2011,50 @@ def test_set_fill_style(modeler: Modeler):
...
+        Point3D([1 / 2, 1 / 2, 0.0], unit=unit),
+        UNITVECTOR3D_X,
+        UNITVECTOR3D_Y,
+    )
+    box_plane = Sketch(plane)
+    box_plane.box(Point2D([0.0, 0.0]), width=1 * unit, height=1 * unit)
+    box = design.extrude_sketch("Block", box_plane, 1 * unit)
+
+    # Default body color is if it is not set on server side.
+    assert box.color == "#000000"
```
jonahrb, line 2031: is the default color not that spaceclaim green?

umutsoysalansys: @jonahrb, SC GetColor returns null on default case, so its just defaulted to this APIServer.

src/ansys/geometry/core/designer/body.py
```diff
@@ -757,7 +757,7 @@ def fill_style(self, value: FillStyle):  # noqa: D102
     def color(self) -> str:  # noqa: D102
         """Get the current color of the body."""
         if self._color is None:
-            if self._grpc_client.backend_version < (25, 1, 0):
+            if self._grpc_client.backend_version < (25, 1, 0): # pragma: no cover
```
umutsoysalansys: why pragma: no cover here? @RobPasMue

RobPasMue: We won't be able to check coverage here since this is for older versions and we are not actively testing against them

src/ansys/geometry/core/designer/body.py
```diff
@@ -962,9 +971,9 @@ def set_fill_style(  # noqa: D102
     @protect_grpc
     @check_input_types
     @min_backend_version(25, 1, 0)
-    def set_color(self, color: Union[str, Tuple[float, float, float], mcolors.Colormap]) -> None:
```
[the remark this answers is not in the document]
RobPasMue: I meant to "use" mcolors to extract the hex. Nonetheless, it doesn't make sense to pass a Colormap @umutsoysalansys. You use colormap to extract the hex value or rgb(a) value of a color. But you have to understand Colormap as a "dictionary". We are only expecting a color - a single color. That's why accepting the color as a string, HEX or as RGB(A) is enough. And you use colormap to convert any of them to HEX - which is what you were already doing...

Bottom line, I just fixed your type hint because it didn't make sense. internally it was fine.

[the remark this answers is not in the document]
RobPasMue: PR is ready but as long as the unstable image is not passing we can't verify it... @smereu and @b-matteo are aware of this. We need to fix that first. We can't merge it otherwise.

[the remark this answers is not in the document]
RobPasMue: That's because the null value is not implemented on the ``windows-latest`` image but on the ``windows-latest-unstable``... however, that image is currently broken. That's why we have to wait for the team to fix the image first.

RobPasMue (approved):
LGTM!

PR #1375, opened 2024-08-27
feat: add ``modeler.exit()`` method

RobPasMue:
## Description
We have the ``modeler.close()`` method but in order to be in line with the other PyAnsys libraries, we are also including ``modeler.exit()`` now. They will effectively perform the same calls.

## Issue linked
Related to https://github.com/ansys/pyansys-geometry/discussions/1374

## Checklist

### DIFF
--- a/src/ansys/geometry/core/modeler.py
+++ b/src/ansys/geometry/core/modeler.py
@@ -221,7 +221,7 @@ class Modeler:
 
     def exit(self) -> None:
         """Access the client's close method.
-        
+
         Notes
         -----
         This method is calling the same method as
--- a/src/ansys/geometry/core/modeler.py
+++ b/src/ansys/geometry/core/modeler.py
@@ -217,7 +217,7 @@ class Modeler:
 
     def close(self) -> None:
         """Access the client's close method."""
-        return self.client.close()
+        self.client.close()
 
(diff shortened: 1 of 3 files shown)


src/ansys/geometry/core/modeler.py
```diff
@@ -219,6 +219,16 @@ def close(self) -> None:
...
 
+    def exit(self) -> None:
+        """Access the client's close method.
+
+        Notes
+        -----
+        This method is calling the same method as
+        :func:`close() <ansys.geometry.core.modeler.Modeler.close>`.
+        """
+        return self.close()
```
AlejandroFernandezLuces, line 230: Why the return? If I'm right, in `client.close()` it doesn't return anything. If the return is needed, I would document it if there is any info being returned.

https://github.com/ansys/pyansys-geometry/blob/310cde57cf4b78e056486e25206f5e15c99216cc/src/ansys/geometry/core/connection/client.py#L278-L301

AlejandroFernandezLuces (dismissed):
LGTM, minor comment.

src/ansys/geometry/core/modeler.py
RobPasMue: Sure, I can change that

PR #1386, opened 2024-08-29
feat: default logs folder on Geometry Service started by Python at PUBLIC (Windows)

### LINKED ISSUE #1377: Perform logging in AppData folder rather than the products folder
Aha so that was it! I'm glad you figured it out. No, I think that's the only folder you need to change the permissions on. Mostly because the logging takes place there with the Geometry Service. BTW, your suggestion makes total sense so I will incorporate it soon! I really like the idea of having it in ``C:\Users\%username%\AppData\Roaming\Ansys\v242``. Let me work on that!

Also, if my initial response helped you out to install the Geometry Service (which is the original request of the discussion), I would kindly ask you to click on "Mark as answer" on the first comment of this thread so that other users can benefit from it!

I appreciate all your time and effort to get it working @dkowollik! Your experience will definitely help other users onboard PyAnsys Geometry!

_Originally posted by @RobPasMue in https://github.com/ansys/pyansys-geometry/discussions/1374#discussioncomment-10464653_

RobPasMue:
## Description
Whenever we launch the Geometry Service locally (on Windows), we will make use of the APPDATA directory for Ansys products. Users should always have write permissions to it, and it will prevent from having some common issues like https://github.com/ansys/pyansys-geometry/discussions/1374#discussioncomment-10464047

## Issue linked
Closes #1377 

## Checklist

### DIFF
--- a/doc/changelog.d/1386.added.md
+++ b/doc/changelog.d/1386.added.md
@@ -1,1 +1,1 @@
-logs on Geometry Service started by Python client always on specific dir at APPDATA (Windows)
+default logs folder on Geometry Service started by Python at PUBLIC (Windows)
--- a/src/ansys/geometry/core/connection/launcher.py
+++ b/src/ansys/geometry/core/connection/launcher.py
@@ -578,16 +578,10 @@ def launch_modeler_with_geometry_service(
         )
 
     # If we are in a Windows environment, we are going to write down the server
-    # logs in the %APPDATA%/Ansys/v{product_version}/GeometryService folder.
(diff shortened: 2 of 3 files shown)


MaxJPRey (commented):
Should we raise an exception in case the user does not have permissions for APPDATA?
It happens quite often.

RobPasMue:
> Should we raise an exception in case the user does not have permissions for APPDATA? It happens quite often.

Well.. That's a good point. We could check whether a user has write access to the folder even before starting (in case it is provided). And if not, we throw an error. Let me check that anyway :)

PR #1420, opened 2024-09-10
fix: change logo link so that it renders properly on PyPI

RobPasMue:
As title says - otherwise logo does not render on PyPI. See https://pypi.org/project/ansys-geometry-core/0.7.1/

See PyMechanical, where this is solved: https://pypi.org/project/ansys-mechanical-core/0.11.7/


RobPasMue:
@jorgepiloto - just FYI

RobPasMue:
Pinging @ansys/pyansys-core for visibility on other projects

PR #1478, opened 2024-10-14
fix: static search options

jorgepiloto:
## Description

With the new release of [ansys-sphinx-theme](https://pypi.org/project/ansys-sphinx-theme/), the name of the options need to be the same as the ones in [Fuse.js options](https://www.fusejs.io/api/options.html). This pull-request ensures that.

### DIFF
--- a/doc/changelog.d/1478.fixed.md
+++ b/doc/changelog.d/1478.fixed.md
@@ -1,0 +1,1 @@
+static search options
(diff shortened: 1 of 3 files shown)


RobPasMue (approved):
Thanks for keeping an eye on this @jorgepiloto - I would have never guessed.

RobPasMue:
Funny enough though... the released docs (which are using the latest version) still work properly with the expected min chatacters. Are we sure about this?

RobPasMue:
Oh wait... you mean with release 1.1.4 which was done today @jorgepiloto?

RobPasMue (approved):

PR #1483, opened 2024-10-16
fix: doc style action

RobPasMue:
Validate that https://github.com/ansys/actions/pull/614 works


RobPasMue:
Closing since it has been validated

PR #1485, opened 2024-10-16
docs: avoid having a drop down in the top navigation bar

jorgepiloto:
## Description

> This is completely optional, so feel free to discard and close.

This pull-request gets rid of the drop-down in the top navigation bar, accommodating all the headers before the drop-down comes in.


RobPasMue (approved):
Aha! Thanks @jorgepiloto - looks really nice! Merging

PR #1584, opened 2024-12-12
fix: vtk/pyvista issues

RobPasMue:
Reverting to Python 3.12 and vtk 9.3.1... something is going on with the self-hosted runner


RobPasMue:
Let's see if reverting it makes it work

RobPasMue:
Okay, we have to revert to Python 3.12 for testing and documentation due to our self-hosted runners... we will get to solve this eventually.

PR #1648, opened 2025-01-16
feat: rename object

jacobrkerstetter:
## Description
Implemented the call to rename object and added the appropriate unit test
Reinstated the _clear_cached_bodies method to keep client and server in sync

## Checklist


src/ansys/geometry/core/designer/geometry_commands.py
```diff
@@ -479,3 +481,20 @@ def extrude_edges_up_to(
...
+    def rename_object(self, selection: List[EntityIdentifier], name: str) -> bool:
+        """Rename an object.
+
+        selection : List[str]
+            Selection of the object to rename.
+        name : str
+            New name for the object.
+        bool
+            ``True`` when successful, ``False`` when failed.
+        """
```
RobPasMue, line 496: Make sure the docstring is properly formatted. You need the "Parameters" title, the "Returns" section etc. Have a look at other methods

src/ansys/geometry/core/designer/geometry_commands.py
```diff
@@ -479,3 +481,20 @@ def extrude_edges_up_to(
         else:
             self._grpc_client.log.info("Failed to extrude edges.")
             return []
+
+    @protect_grpc
+    @min_backend_version(25, 2, 0)
+    def rename_object(self, selection: List[EntityIdentifier], name: str) -> bool:
```
RobPasMue, line 487: Inputs should not be a gRPC object. We need to solve this. You should retrieve the gRPC ID from the PyAnsys Geometry objects. You might have to follow an approach similar to the one here where we are looking for an attribute called "_parent_component"

https://github.com/ansys/pyansys-geometry/blob/f37c6f30beaf27f84d98eb55ecb9343285ec1899/src/ansys/geometry/core/misc/checks.py#L53-L66

In your case you will have to look for "id" or even "_grpc_id". Some objects store it.

Also, I don't get something... we are renaming (with the same name) a list of objects? I would expect it only to be a single input, not a list

RobPasMue (changes requested):
Requesting a few changes..

src/ansys/geometry/core/designer/geometry_commands.py
RobPasMue: I meant mostly on the Python side. All our objects (or most of them) have an identifier. We should be passing in Python objects and extracting the Entity Identifier from them. Maybe @jonahrb can guide you on this. Otherwise we can sync in the dev meeting

tests/integration/test_geometry_commands.py
```diff
@@ -252,3 +252,24 @@ def test_extrude_edges_and_up_to(modeler: Modeler):
     assert created_bodies[0].faces[3].area.m == pytest.approx(
         Quantity(6, UNITS.m**2).m, rel=1e-6, abs=1e-8
     )
+
+
+def test_rename_object(modeler: Modeler):
```
umutsoysalansys, line 257: hi @jacobrkerstetter , does this object renaming also works for geometries other than bodies?

umutsoysalansys: asking because we already had get name and set name methods for body. but I guess this one looks more general and apply things other than bodies, is that right?

src/ansys/geometry/core/designer/geometry_commands.py
jacobrkerstetter: Ah i understand that, i am fixing it as we speak. I meant on the single input vs. list.... i'm not too sure about that part

tests/integration/test_geometry_commands.py
jacobrkerstetter: it should also work on components. to my understanding its anything with a "name" property

src/ansys/geometry/core/designer/geometry_commands.py
jonahrb: @jacobrkerstetter  Yes i would say remove the "EntityIdentifier" and instead allow a "Body" or "Component" , also add some unit tests for Components :) pm me on teams if you have questions

src/ansys/geometry/core/designer/geometry_commands.py
```diff
@@ -496,6 +498,28 @@ def extrude_edges_up_to(
 
     @protect_grpc
     @min_backend_version(25, 2, 0)
+    def rename_object(self, selection: List[EntityIdentifier], name: str) -> bool:
```
RobPasMue, line 501: Type hint is not correct. Also, we should be passing in the ``Body``, ``Component``, ``Face``, ``Edge`` objects rather than a list of strings. Internally (within this method) we should extract the IDs. It might imply creating a generic method which we can use elsewhere

RobPasMue (changes requested):

src/ansys/geometry/core/designer/geometry_commands.py
```diff
@@ -524,11 +553,6 @@ def create_linear_pattern(
             How many times the pattern repeats in the y direction.
         pitch_y : Real, default: None
             The spacing between each pattern member in the y direction.
```
RobPasMue, line 555:
```suggestion
            The spacing between each pattern member in the y direction.

            Returns
            -------
            bool
                ``True`` when successful, ``False`` when failed.
```

src/ansys/geometry/core/designer/geometry_commands.py
```diff
@@ -525,10 +554,10 @@ def create_linear_pattern(
...
 
-        Returns
-        -------
-        bool
-            ``True`` when successful, ``False`` when failed.
+            Returns
+            -------
+            bool
+                ``True`` when successful, ``False`` when failed.
```
RobPasMue:
```suggestion
        Returns
        -------
        bool
            ``True`` when successful, ``False`` when failed.
```

src/ansys/geometry/core/designer/geometry_commands.py
```diff
@@ -527,8 +556,8 @@ def create_linear_pattern(
 
         Returns
         -------
-        bool
-            ``True`` when successful, ``False`` when failed.
+            bool
+                ``True`` when successful, ``False`` when failed.
         """
```
RobPasMue, line 561:
```suggestion
        bool
            ``True`` when successful, ``False`` when failed.
        """
```

RobPasMue (approved):
LGTM!

PR #1649, opened 2025-01-16
matrix operations

umutsoysalansys:
## Description

## Issue linked

## Checklist

### DIFF
--- a/src/ansys/geometry/core/math/matrix.py
+++ b/src/ansys/geometry/core/math/matrix.py
@@ -26,6 +26,8 @@ from typing import Union
 from beartype import beartype as check_input_types
 import numpy as np
 
+from ansys.geometry.core.math.frame import Frame
+from ansys.geometry.core.math.vector import Vector3D
 from ansys.geometry.core.misc.checks import check_ndarray_is_float_int
 from ansys.geometry.core.typing import Real, RealSequence
 
@@ -129,3 +131,36 @@ class Matrix44(Matrix):
             raise ValueError("Matrix44 should only be a 2D array of shape (4,4).")
 
         return obj
+
+    @staticmethod
+    def create_matrix_from_rotation(direction_x: Vector3D, direction_y: Vector3D) -> "Matrix44":
+        """Matrix44 helper method -- create_matrix_from_rotation."""
+        matrix = Matrix44(
+            [
+                [direction_x.x, direction_x.y, direction_x.z, 0],
+                [direction_y.x, direction_y.y, direction_y.z, 0],
+                [0, 0, 1, 0],
+                [0, 0, 0, 1],
+            ]
+        )
+        return matrix
+
+    @staticmethod
+    def create_matrix_from_translation(origin: Vector3D) -> "Matrix44":
+        """Matrix44 helper method -- create_matrix_from_translation."""
+        matrix = Matrix44(
+            [
+                [1, 0, 0, origin.x],
+                [0, 1, 0, origin.y],
+                [0, 0, 1, origin.z],
+                [0, 0, 0, 1],
+            ]
+        )
+        return matrix
+
+    @staticmethod
+    def create_matrix_from_mapping(frame: Frame) -> "Matrix44":
+        """Matrix44 helper method -- create_matrix_from_mapping."""
+        translation_matrix = Matrix44.create_matrix_from_translation(frame.origin)
+        rotation_matrix = Matrix44.create_matrix_from_rotation(frame.direction_x, frame.direction_y)
+        return translation_matrix * rotation_matrix
--- a/src/ansys/geometry/core/math/vector.py
+++ b/src/ansys/geometry/core/math/vector.py
@@ -28,6 +28,7 @@ from beartype import beartype as check_input_types
 import numpy as np
 from pint import Quantity
 
+from ansys.geometry.core import math
 from ansys.geometry.core.math.matrix import Matrix44
 from ansys.geometry.core.math.point import Point2D, Point3D
 from ansys.geometry.core.misc.accuracy import Accuracy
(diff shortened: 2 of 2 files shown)


RobPasMue:
@umutsoysalansys - please modify the PR description according to the guidelines whenever you get the chance. This allows us to properly maintain the repository and easily determine features and implementations

tests/test_math.py
```diff
@@ -752,6 +752,35 @@ def test_matrix_44():
...
+    expected_matrix = Matrix44([[1, 0, 0, 0], [0, 1, 0, 0], [0, 0, 1, 0], [0, 0, 0, 1]])
+    rotation_matrix = Matrix44.create_matrix_from_rotation(direction_x, direction_y)
+    assert np.array_equal(expected_matrix, rotation_matrix)
+
+
+def test_create_matrix_from_translation():
+    origin = Vector3D([1, 2, 3])
+
+    expected_matrix = Matrix44([[1, 0, 0, 1], [0, 1, 0, 2], [0, 0, 1, 3], [0, 0, 0, 1]])
+    translation_matrix = Matrix44.create_matrix_from_translation(origin)
```
RobPasMue, line 768: Origin looks odd.. shouldn't it be a Point3D?

umutsoysalansys: might as well, thanks you've got a point there:)

src/ansys/geometry/core/math/matrix.py
```diff
@@ -129,3 +134,40 @@ def __new__(cls, input: np.ndarray | RealSequence | Matrix = DEFAULT_MATRIX44):
...
+        """Matrix44 helper method -- create_matrix_from_rotation."""
+        matrix = cls(
+            [
+                [direction_x.x, direction_x.y, direction_x.z, 0],
+                [direction_y.x, direction_y.y, direction_y.z, 0],
+                [0, 0, 1, 0],
+                [0, 0, 0, 1],
+            ]
+        )
+        return matrix
```
RobPasMue, line 151: Why can't there be a Z direction? A rotation is typically defined in any direction, especially in 3D space. You are assuming a planar rotation with this method (with a frozen Z axis) - meaning you can only change the X Y  ref vectors. 

Rotations in space are defined with either quaternions or euler angles (both the angle values and the order of rotation). Please modify accordingly.

src/ansys/geometry/core/math/matrix.py
```diff
@@ -129,3 +134,40 @@ def __new__(cls, input: np.ndarray | RealSequence | Matrix = DEFAULT_MATRIX44):
...
+
+    @classmethod
+    def create_matrix_from_translation(cls, origin: "Point3D") -> "Matrix44":
+        """Matrix44 helper method -- create_matrix_from_translation."""
```
RobPasMue, line 155: A translation should receive as input a Vector in 3D space, not a Point. You have to modify the input and the implementation accordingly.

src/ansys/geometry/core/math/matrix.py
```diff
@@ -129,3 +134,40 @@ def __new__(cls, input: np.ndarray | RealSequence | Matrix = DEFAULT_MATRIX44):
...
+        )
+        return matrix
+
+    @classmethod
+    def create_matrix_from_mapping(cls, frame: "Frame") -> "Matrix44":  # noqa
+        """Matrix44 helper method -- create_matrix_from_mapping."""
+        translation_matrix = cls.create_matrix_from_translation(frame.origin)
+        rotation_matrix = cls.create_matrix_from_rotation(frame.direction_x, frame.direction_y)
+
+        return translation_matrix * rotation_matrix
```
RobPasMue, line 173: This one is interesting but honestly... it's easier to be done manually.

If you want to move from Frame A to Frame B, the rotation is defined by the following rotation matrix (where **i**, **j** and **k** are the axis of Frame A and **i'**, **j'** and **k'** the directions of Frame B

image

http://kwon3d.com/theory/transform/transform.html

On top of this, you want to add the frame translation (i.e. Origin of Frame B - Origin of Frame A = Translation vector).
Let's call it our **t** vector based on the following notation:

image

So, you can actually build the full matrix yourself. I would recommend to do so rather than to chain together these methods. Seems easier to me at least.

src/ansys/geometry/core/math/vector.py
```diff
@@ -265,6 +266,21 @@ class representing the second point.
         """
         return Vector3D(point_b - point_a)
 
+    @staticmethod
+    def rotate_vector(_this: "Vector3D", vector: "Vector3D", angle: float) -> "Vector3D":
```
RobPasMue, line 270: Do not call it ``_this``. That doesn't make sense. A static method is not related to any instance. So there is no concept of ``this``, ``self`` or anything like that. Rename to ``vec_a`` and ``vec_b`` for example.

src/ansys/geometry/core/math/vector.py
```diff
@@ -265,6 +266,21 @@ class representing the second point.
...
+            raise Exception("Invalid vector operation.")
+
+        angle_between = vector.get_angle_between(_this)
+        parallel = Vector3D(
+            vector.x * angle_between, vector.y * angle_between, vector.z * angle_between
+        )
+
+        perpendicular1 = vector - parallel
+        perpendicular2 = _this.cross(perpendicular1)
+        return parallel + perpendicular1 * math.cos(angle) + perpendicular2 * math.sin(angle)
```
RobPasMue, line 282: Watch out... you are performing many assumptions here.

1) Why are you multiplying the coordinates against the angle alone? If you are assuming an infinitesimal angle (i.e. very small) and you are approximating the sine to the angle value... you could have a point. But you cannot assume the angle is small by default. Please revisit.
2) What's the reference for this algorithm? Is there some text-book where it is showing up? I can't replicate your math sadly. Please share the link. Also, we need to improve the in-line comments and explanations

RobPasMue (changes requested):
Requesting changes. Also, please generate proper docstrings for all the methods @umutsoysalansys - they are very undocumented as they are right now..

src/ansys/geometry/core/math/vector.py
umutsoysalansys: @RobPasMue makes sense, let me rework on it

RobPasMue: Sure! Feel free to reach out if you want to discuss anything =)

src/ansys/geometry/core/math/vector.py
umutsoysalansys: taking this out of scope of this PR

src/ansys/geometry/core/math/matrix.py
umutsoysalansys: @jrwest440  do you have any comment on this?

PR #1666, opened 2025-01-20
feat: revolve faces a set distance, up to another object, or by a helix

jonahrb:
## Description
Revolve faces around an axis.

## Checklist

### DIFF
--- a/src/ansys/geometry/core/connection/conversions.py
+++ b/src/ansys/geometry/core/connection/conversions.py
@@ -662,3 +662,21 @@ def trimmed_surface_to_grpc_trimmed_surface(
         v_min=trimmed_surface.box_uv.interval_v.start,
         v_max=trimmed_surface.box_uv.interval_v.end,
     )
+
+
+def line_to_grpc_line(line: Line) -> GRPCLine:
+    """Convert a ``Line`` to a line gRPC message.
+
+    Parameters
+    ----------
+    line : Line
+        Line to convert.
+
+    Returns
+    -------
+    GRPCLine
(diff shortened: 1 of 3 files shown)


src/ansys/geometry/core/connection/conversions.py
```diff
@@ -662,3 +662,21 @@ def trimmed_surface_to_grpc_trimmed_surface(
         v_min=trimmed_surface.box_uv.interval_v.start,
         v_max=trimmed_surface.box_uv.interval_v.end,
     )
+
+
+def line_to_grpc_line(line: Line) -> GRPCLine:
```
RobPasMue, line 667: Can't we use 
```
def curve_to_grpc_curve(curve: Curve) -> GRPCCurve:
````
for this purpose? Why do we have a separate GRPCLine message? Can't we handle it as a GRPCCurve?

RobPasMue (approved):
Overall, LGTM - just a small comment

RobPasMue:
Looks like there are some issues with the helix test ref values. Have a look at them whenever you get the chance @jonahrb

src/ansys/geometry/core/connection/conversions.py
jonahrb: @RobPasMue It's true we could, but it could be confused on the ApiServer side that any type of curve would be accepted for the axis argument, but only a Line will work, so I'm going to leave it explicit like this

RobPasMue: Okay understood - thanks @jonahrb

PR #1669, opened 2025-01-21
feat: add split body and tests

david-gorman:
## Description
**Added functionality to split body by planes, splitters, or faces.**

## Issue linked

## Checklist


RobPasMue:
Hi @david-gorman - could you solve the existing conflicts between your branch and the blitz branch? I will start reviewing nonetheless :)

src/ansys/geometry/core/designer/geometry_commands.py
```diff
@@ -842,3 +846,68 @@ def update_fill_pattern(
...
+        if plane is not None:
+            plane_item = plane_to_grpc_plane(plane)
+
+        slicer_items = None
+        if slicers is not None:
+            slicer_items = [slicer._grpc_id for slicer in slicers]
+
+        face_items = None
+        if faces is not None:
+            face_items = [face._grpc_id for face in faces]
```
RobPasMue, line 897: Before iterating on the method inputs we should be performing some type checking since we are using loose type hints (i.e. "Face" instead of Face + checker decorator)

RobPasMue, line 897: Before iterating on the method inputs we should be performing some type checking since we are using loose type hints (i.e. "Face" instead of Face + checker decorator)

RobPasMue (approved):
Apart from my comments - LGTM @david-gorman - thanks for the implementation!

src/ansys/geometry/core/designer/geometry_commands.py
david-gorman: @RobPasMue  I add a check_input_types decorator to this method.  I think that's what you meant.  Please let me if not!

RobPasMue: Yeah that's exactly what's not going to work... for the decorator to work, it is necessary that the type hints are actually imported on the module (i.e. outside of the ``if TYPE_CHECKING`` conditional).

Since this is not the case, you have to verify them inside the method. You can use any of the checker functions available in https://github.com/ansys/pyansys-geometry/blob/main/src/ansys/geometry/core/misc/checks.py

RobPasMue (approved):

PR #1671, opened 2025-01-21
feat: enable get/set persistent ids for stride import/export

MikeJanes:
## Description

## Issue linked

## Checklist


src/ansys/geometry/core/misc/auxiliary.py
```diff
@@ -140,6 +140,26 @@ def __traverse_all_bodies(comp: Union["Design", "Component"]) -> list["Body"]:
...
+    list[Body]
+        List of Body objects.
+
+    Notes
+    -----
+    This method takes a design and gets the corresponding ``Body`` objects.
+    """
+    return __traverse_all_bodies(design)
+
+
```
umutsoysalansys, line 162: @mlkaplan36 just out of curiosity, are there any cases where design.bodies is not working well enough?

MikeJanes: @umutsoysalansys I believe that only gives the bodies directly under that design/component. I want all the bodies, including bodies in sub components.

src/ansys/geometry/core/misc/unsupported.py
```diff
@@ -0,0 +1,276 @@
...
+            or id_type not in self.__id_map
+        ):
+            self.__fill_imported_id_map(id_type)
+        moniker = self.__id_map[id_type].get(import_id, None)
+        return moniker
+
+    @protect_grpc
+    @min_backend_version(25, 2, 0)
+    def set_export_id(
+        self, moniker: "EntityIdentifier", id_type: "PersistentIdType", value: "str"
```
jonahrb, line 160: since this is a public-facing method, moniker should be a string instead of an EntityIdentifier - so users don't have to know about and use grpc proto message objects. Some objects like `Body` already have a `Body._grpc_id` property that turns its string id into an EntityIdentifier

src/ansys/geometry/core/misc/unsupported.py
```diff
@@ -0,0 +1,276 @@
...
+
+    Parameters
+    ----------
+    grpc_client : GrpcClient
+        gRPC client to use for the geometry commands.
+    """
+
+    @protect_grpc
+    def __init__(self, grpc_client: GrpcClient, modeler):
+        """Initialize an instance of the ``GeometryCommands`` class."""
```
jonahrb, line 62:
```suggestion
        """Initialize an instance of the ``UnsupportedCommands`` class."""
```

src/ansys/geometry/core/misc/unsupported.py
```diff
@@ -0,0 +1,276 @@
...
+    @min_backend_version(25, 2, 0)
+    def __fill_imported_id_map(self, id_type: "PersistentIdType") -> None:
+        """Populate the persistent id map for caching.
+
+        Parameters
+        ----------
+        id_type : PersistentIdType
+            type of id
+
+        Returns
```
jonahrb, line 79: You can remove "Returns" sections from void methods

jonahrb (approved):

src/ansys/geometry/core/misc/unsupported.py
```diff
@@ -0,0 +1,268 @@
+# Copyright (C) 2023 - 2025 ANSYS, Inc. and/or its affiliates.
```
RobPasMue, line 1: This entire module shouldn't be here... It should go inside the ``tools`` subpackage IMO.

tests/integration/test_design_import.py
```diff
@@ -170,6 +171,34 @@ def test_open_file(modeler: Modeler, tmp_path_factory: pytest.TempPathFactory):
...
+    bodies2 = modeler.unsupported.get_body_occurrences_from_import_id(
+        "2", PersistentIdType.PRIME_ID
+    )
+
+    # requires a change to core service, uncomment on next core service update
+    # assert base_body.id in [b.id for b in bodies1]
+    # assert wheel_body.id in [b.id for b in bodies2]
+    assert base_body.id not in [b.id for b in bodies1]
+    assert wheel_body.id not in [b.id for b in bodies2]
+
```
RobPasMue, line 192: This is going to prevent the update of the Docker images unless we do it right away.... we are checking that the new images are not breaking the tests before promoting them. This is done on ``main`` rather than on ``blitz``. And we are safe because of that but nonetheless, it is going to be problematic on the workflows

RobPasMue (commented):
Let's hold off merging if possible next time until I have had a look to the PRs. There are some changes requested on my side for this module. Let's handle them when possible @MikeJanes.

Also, the docstrings are not in-line with what the rest of the library is performing. I am talking mostly about the ones inside the ``unsupported`` module. Let's revisit them.

The min_backend_version decorator has been overused as well. All private methods "__" shouldn't have them, because you should have checked it in the calling method already (causing an unnecessary reduction in performance).

PR #1675, opened 2025-01-22
feat: implementation of NURBS curves

### LINKED ISSUE #46: Implement the Spline object
We are definitely not going to implement the Spline logic on our own. We should look into other libraries to see who has it implemented and wrap around it.

RobPasMue:
## Description
Implementation of NURBS-based curves on Python side.

## Issue linked
Closes #46

## Checklist


RobPasMue:
We are still pending to make the link with the server side, for extrusion paths for example... requires syncing with @mlkaplan36

RobPasMue:
Link between client and server will be done later. Partially closing #46 consequently

src/ansys/geometry/core/shapes/curves/nurbs.py
```diff
@@ -0,0 +1,310 @@
...
+        The parameter is defined in the interval [0, 1] by default. Information
+        is provided about the parameter type and form.
+
+        Returns
+        -------
+        Parameterization
+            Information about how the NURBS curve is parameterized.
+        """
+        return Parameterization(
+            ParamType.OTHER,
```
jonahrb, line 157: The `ParamType` for a NURBS curve can OPEN, CLOSED, or PERIODIC based on the actual shape of the curve. If the library provides this information then great, if not maybe just leave it as OTHER.

jonahrb (approved):

mlkaplan36 (approved):

src/ansys/geometry/core/shapes/curves/nurbs.py
RobPasMue: Ah good point. I think I can determine that from the library. Let's see :)

RobPasMue: Hard to determine. The library does not implement these functionalities. I know that the NURBS curve has a way to determine if a the NURBS curve is periodic or not but I do not have access to the book or the algorithm. We can check offline @jonahrb - merging as is and we can come back to this later

PR #1685, opened 2025-01-24
fix: filter set export id to only CoreService based backends

b-matteo:
## Description
This change is to fix the test_open_file. We used to call set and get export id for whatever backend type we where testing against. Right now those apis are only implemented in CoreService based backends, so we should only use set and get export id against those.
## Checklist


RobPasMue (approved):
LGTM!

PR #1740, opened 2025-02-11
feat: conservative approach to single design per modeler

### LINKED ISSUE #1710: Single design instance per Modeler object
Instead of allowing for multiple designs and all its implications, we should move towards a single design per Modeler mechanism.

RobPasMue:
## Description
A single design instance per Modeler object is handled. If a new design is requested, the old design is closed. We are also being conservative and checking that if you use an object belonging to an old design, we are still checking whether the design is active or not. With #1721 we lost this capability. This implementation makes it more robust

## Issue linked
Closes #1710

## Checklist

### DIFF
--- a/src/ansys/geometry/core/modeler.py
+++ b/src/ansys/geometry/core/modeler.py
@@ -37,8 +37,7 @@ from ansys.geometry.core.connection.backend import ApiVersions, BackendType
 from ansys.geometry.core.connection.client import GrpcClient
 from ansys.geometry.core.connection.defaults import DEFAULT_HOST, DEFAULT_PORT
 from ansys.geometry.core.errors import GeometryRuntimeError, protect_grpc
-from ansys.geometry.core.logger import LOG
-from ansys.geometry.core.misc.checks import check_type, min_backend_version
+from ansys.geometry.core.misc.checks import check_type, deprecated_method, min_backend_version
 from ansys.geometry.core.misc.options import ImportOptions
 from ansys.geometry.core.tools.measurement_tools import MeasurementTools
 from ansys.geometry.core.tools.prepare_tools import PrepareTools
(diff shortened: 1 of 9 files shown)


smereu:
@RobPasMue I played around with a few things with Discovery and core services and I didn't encounter any issue. Looks good!

I have a few comments that are minor and can probably be addressed (if needed) after the PR is merged
1) I see that modeler.designs[] is marked as obsolete but is there way to know that it is the case? I only know because I looked at the code :-)
2) When you close a design, its property is_active becomes False.  There is also a property is_alive (as it derives from Component) that is still set to True.  Maybe it will disappear with the next round of refactoring. If not, we can set it to False as well
3) I see that there is a call to _activate(). Hopefully no code path will hit it, but I wonder if it can/should be removed
4) The decorator @ensure_design_is_active is used in many places.  As we do the next round of refactoring maybe we can do something like that internally (since we won't cache anything) so we remove the need to have it almost everywhere

RobPasMue:
> 1. I see that modeler.designs[] is marked as obsolete but is there way to know that it is the case? I only know because I looked at the code :-)

The user will get a warning in their logs whenever they perform a call to ``designs``. It's up to the user to read the logs and see deprecated methods from there onwards :)

> 2. When you close a design, its property is_active becomes False.  There is also a property is_alive (as it derives from Component) that is still set to True.  Maybe it will disappear with the next round of refactoring. If not, we can set it to False as well

Hmm the ``is_alive`` boolean was used for different purposes (i.e. to determine  if a body/component has been deleted or not). All the modeling operations should have the ``@ensure_design_is_Active`` decorator so we should be safe there. If a user still wants to query an old component for info such as... it's name? They could still do it. And I agree that maybe it will just get removed in the next refactor. If it doesn't we can decide then.

> 3. I see that there is a call to _activate(). Hopefully no code path will hit it, but I wonder if it can/should be removed

There are 2 calls to it in fact. But the method is still needed so that the client side still knows that a design is active. 

> 4. The decorator @ensure_design_is_active is used in many places.  As we do the next round of refactoring maybe we can do something like that internally (since we won't cache anything) so we remove the need to have it almost everywhere

That decorator is still needed IMO as long as the caching still exists. But I don't envision it disappearing. It depends on the depth of the caching we will do honestly. We can discuss this later.

src/ansys/geometry/core/modeler.py
```diff
@@ -120,15 +119,15 @@ def __init__(
             backend_type=backend_type,
         )
 
-        # Maintaining references to all designs within the modeler workspace
-        self._designs: dict[str, "Design"] = {}
+        # Single design for the Modeler
+        self._design: Optional["Design"] = None
 
         # Initialize the RepairTools - Not available on Linux
```
jrwest440, line 125: @RobPasMue IS this block still true?  I see it reference repair tools -- which are OK now.  Are measurement tools still not in place?

jrwest440 (approved):

src/ansys/geometry/core/modeler.py
RobPasMue: AFAIK they aren't. They are the only remaining ones.

PR #1783, opened 2025-02-27
build: temporary workaround for using trusted publisher approach

RobPasMue:
## Description
Following issues with trusted publisher listed in https://github.com/ansys/actions/issues/687 - we need to limit the flit version used to build the project. This is a temporary workaround until we properly fix this issue.

## Issue linked
None

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["flit_core >=3.2,<4"]
+requires = ["flit_core >=3.2,<3.11"]
 build-backend = "flit_core.buildapi"
 
 [project]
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["flit_core >=3.2,<4"]
+requires = ["flit_core >=3.2,<3.11"]
```
RobPasMue, line 2:
```suggestion
requires = ["flit_core >=3.2,<3.11"] # THIS SHOULD BE REVERTED TO '["flit_core >=3.2,<4"]'
```

PR #1851, opened 2025-03-19
feat: Remove rounds

smereu:
## Description
Exposes the ability to remove rounds. The round faces to be removed are input to the method.  The functionality was available in previous versions of the API server but CoreServices probably did not implement it until recently. I am using min version 25, 2 decorator but I can change it as needed.

## Issue linked

## Checklist


src/ansys/geometry/core/tools/prepare_tools.py
```diff
@@ -168,6 +169,37 @@ def extract_volume_from_edge_loops(
...
+        check_type_all_elements_in_iterable(faces, Face)
+
+        parent_design = get_design_from_face(faces[0])
+
+        response = self._prepare_stub.RemoveRounds(RemoveRoundsRequest(faces=faces))
+
+        if response.result:
+            parent_design._update_design_inplace()
+        else:
+            self._grpc_client.log.info("Failed to remove rounds...")
```
RobPasMue, line 201: Should we return the response status (i.e. boolean)? I think other tools are doing so.

src/ansys/geometry/core/tools/prepare_tools.py
```diff
@@ -168,6 +169,37 @@ def extract_volume_from_edge_loops(
             self._grpc_client.log.info("Failed to extract volume from edge loops...")
             return []
 
+    @protect_grpc
+    @min_backend_version(25, 2, 0)
+    def remove_rounds(self, faces: list["Face"]):
```
RobPasMue, line 174: Whether we return the response status (boolean) or nothing (None) let's add a typehint. And adapt the docstring if we return the boolean.

RobPasMue (approved):
Implementation looks perfect. Just left some potential enhancements. I see the tests are failing but once that's in we should be good to merge! Thanks @smereu

tests/integration/test_prepare_tools.py
```diff
@@ -56,6 +56,21 @@ def test_volume_extract_from_edge_loops(modeler: Modeler):
     assert len(created_bodies) == 1
 
 
+def test_remove_rounds(modeler: Modeler):
+    """Test a round is removed from the geometry."""
+    from ansys.geometry.core.designer import SurfaceType
+
+    design = modeler.open_file(r"C:\temp\BoxWithRound.scdocx")
```
umutsoysalansys, line 63: @smereu looks like file path needs to be updated.

RobPasMue: Very good catch @umutsoysalansys!! I completely missed that one

RobPasMue:
@smereu - tests were failing because the proto API was expecting "selection" and not "faces". I just modified it and committed the change. Tests should pass after this change I guess.

https://github.com/ansys/pyansys-geometry/pull/1851/commits/17c1acc6821fde28d75b723283ce256d43f39a72

BTW - which is the service version in which this API was implemented? Since CoreService is not available earlier than 25R2, it is not a problem. We can probably add the first version in which this API was added

RobPasMue:
Okay so it seems that the API was available as of 24.1. So we do not need to add the decorator in that case. I will take over a few changes in this PR to align it to other prepare tools (typehints & docstrings). I want to get it in before we merge blitz into main.

PR #1873, opened 2025-03-27
feat: Logo detection

smereu:
## Description
Exposes logo detection and removal. Logos is exposed as a problem area defined by a list of faces. There is no ability to distinguish multiple logos if present so we expose a single problem area. The problem area has a fix method to remove the logo. There is a find and remove method that allows to remove everything automatically.

## Checklist


smereu:
@RobPasMue My previous branch for logo detection had too many conflicts and I was getting a hard time with it so I created a new one and a new pull request.  There is a small fix pending on the API server then it will be ready for full review

src/ansys/geometry/core/tools/problem_areas.py
```diff
@@ -619,3 +625,45 @@ def fix(self) -> RepairToolMessage:
...
+    """
+
+    def __init__(self, id: str, grpc_client: GrpcClient, face_ids: list[str]):
+        """Initialize a new instance of the logo problem area class."""
+        super().__init__(id, grpc_client)
+
+        self._face_ids = face_ids
+
+    @property
+    def face_ids(self) -> list["str"]:
```
MaxJPRey, line 650: We might not need the quotes around str as it is a built-in type.
```suggestion
    def face_ids(self) -> list[str]:
```

smereu: Fixed. Thx :-)

smereu:
@RobPasMue I added some check for disabling on Linux but I am not sure if it is the correct way to do it. Should I log an issue for it?  Also there is one change in the API server that has to be merged otherwise one of the asserts will fail.

RobPasMue:
> @RobPasMue I added some check for disabling on Linux but I am not sure if it is the correct way to do it. Should I log an issue for it? Also there is one change in the API server that has to be merged otherwise one of the asserts will fail.

Hi @smereu -- I can't see any "skip on Linux" condition on your PR, sadly. That's what's causing the CI/CD to fail I guess.

src/ansys/geometry/core/tools/prepare_tools.py
```diff
@@ -295,3 +297,93 @@ def enhanced_share_topology(
...
+        min_height : real, optional
+            The minimum height when searching for logos
+        max_height: real, optional
+            The minimum height when searching for logos
+
+        Returns
+        -------
+        Logo problem area
+
+            Problem area with logo faces.
```
RobPasMue, line 324:
```suggestion
        LogoProblemArea
            Problem area with logo faces.
```

src/ansys/geometry/core/tools/prepare_tools.py
```diff
@@ -295,3 +297,93 @@ def enhanced_share_topology(
...
+        if bodies:
+            if len(bodies) > 0:
+                check_type_all_elements_in_iterable(bodies, Body)
+
+        body_ids = [] if bodies is None else [body._grpc_id for body in bodies]
+        find_logo_options = FindLogoOptions()
+        if min_height:
+            find_logo_options.min_height = min_height
+        if max_height:
+            find_logo_options.max_height = max_height
```
RobPasMue, line 338: Couple of things:

1. We are not verifying that the inputs are actually numbers. We should do that.
2. I believe we could simplify this to the following:
```suggestion
        find_logo_options = FindLogoOptions(
            min_height=min_height,
            max_height=max_height,
        )
```

If the values are None, gRPC shouldn't convert them into the message anyway.

src/ansys/geometry/core/tools/prepare_tools.py
```diff
@@ -295,3 +297,93 @@ def enhanced_share_topology(
...
+
+        response = self._prepare_stub.FindLogos(
+            FindLogosRequest(bodies=body_ids, options=find_logo_options)
+        )
+
+        face_ids = []
+        for grpc_face in response.logo_faces:
+            face_ids.append(grpc_face.id)
+        return LogoProblemArea(id=response.id, grpc_client=self._grpc_client, face_ids=face_ids)
```
RobPasMue, line 347:
```suggestion
        return LogoProblemArea(id=response.id, grpc_client=self._grpc_client, face_ids=[grpc_face.id for grpc_face in response.logo_faces])
```
This can probably be simplified into this.

src/ansys/geometry/core/tools/prepare_tools.py
```diff
@@ -295,3 +297,93 @@ def enhanced_share_topology(
...
+            if len(bodies) > 0:
+                check_type_all_elements_in_iterable(bodies, Body)
+
+        body_ids = [] if bodies is None else [body._grpc_id for body in bodies]
+        find_logo_options = FindLogoOptions()
+        if min_height:
+            find_logo_options.min_height = min_height
+        if max_height:
+            find_logo_options.max_height = max_height
+
```
RobPasMue, line 384: Same changes as before to be applied here.

src/ansys/geometry/core/tools/prepare_tools.py
```diff
@@ -295,3 +297,93 @@ def enhanced_share_topology(
...
+        bodies : list[Body], optional
+            List of bodies where logos should be detected and removed.
+        min_height : real, optional
+            The minimum height when searching for logos
+        max_height: real, optional
+            The minimum height when searching for logos
+
+        Returns
+        -------
+        Boolean value indicating whether the operation was successful.
```
RobPasMue, line 369:
```suggestion
        Returns
        -------
        bool
            Boolean value indicating whether the operation was successful.
```
The format for docstrings is:

```
Returns
-------
type
    description
```

src/ansys/geometry/core/tools/prepare_tools.py
```diff
@@ -295,3 +297,93 @@ def enhanced_share_topology(
...
+        """
+        from ansys.geometry.core.designer.body import Body
+
+        # Verify inputs
+        if bodies:
+            if len(bodies) > 0:
```
RobPasMue, line 375: These nested ifs, please make them into a single conditional
```suggestion
        if bodies and len(bodies) > 0:
```

src/ansys/geometry/core/tools/prepare_tools.py
```diff
@@ -295,3 +297,93 @@ def enhanced_share_topology(
...
+        -------
+        Logo problem area
+
+            Problem area with logo faces.
+        """
+        from ansys.geometry.core.designer.body import Body
+
+        # Verify inputs
+        if bodies:
+            if len(bodies) > 0:
```
RobPasMue, line 330:
```suggestion
        if bodies and len(bodies) > 0:
```
Multiple nested ifs into a single statement

tests/integration/test_prepare_tools.py
```diff
@@ -112,3 +112,36 @@ def test_enhanced_share_topology(modeler: Modeler):
     result = modeler.prepare_tools.enhanced_share_topology(design.bodies, 0.000554167, True)
     assert result.found == 14
     assert result.repaired == 14
+
+
+def test_detect_logos(modeler: Modeler):
+    """Test logos are detected and deleted."""
+    design = modeler.open_file(FILES_DIR / "Part1.SLDPRT")
+    assert len(design.components[0].bodies[2].faces) == 189
```
RobPasMue, line 120: Rather than accessing by index, I would access by body name/component name... this makes the test more reliable for future changes in the way the design is read, for example. This comment applies to both tests.

RobPasMue (changes requested):
Requesting a few changes -- nonetheless, the PR is looking good, thanks @smereu!

src/ansys/geometry/core/tools/prepare_tools.py
smereu: Addressed

smereu:
@RobPasMue I hope I addressed the comments. I will add the quantity input to this method together with all the other ones.
The failing test needs a new build of API server. I could address it now but it will start failing with a new build

src/ansys/geometry/core/tools/prepare_tools.py
```diff
@@ -60,10 +66,11 @@ class PrepareTools:
         Active supporting geometry service instance for design modeling.
     """
 
-    def __init__(self, grpc_client: GrpcClient):
+    def __init__(self, grpc_client: GrpcClient, modeler: "Modeler"):
         """Initialize Prepare Tools class."""
         self._grpc_client = grpc_client
         self._prepare_stub = PrepareToolsStub(self._grpc_client.channel)
+        self._modeler = modeler
```
RobPasMue, line 73: I think there is no need to input the modeler here @smereu - you already have ``_grpc_client`` which also has access to the backend type (``_grpc_client.backend_type``). I would recommend removing the import and usage of Modeler inside the tool if it is only for this purpose.

RobPasMue (commented):
Hi @smereu - changes look good. Thanks for addressing them. I think there are only 2 comments missing to be resolved:

1- I'd rather access by name (inside the test) whenever we want to validate number of faces of a given body. I'd search for the component with the desired name - and for the specific body we want (also by name), and then validate its face number.
2- I don't think we need the modeler inside prepare tools based on the operations you are performing (just retrieving the backend type). That information is available inside the client already so there is no need for the modeler =)

src/ansys/geometry/core/tools/prepare_tools.py
smereu: Applied :-)

src/ansys/geometry/core/tools/prepare_tools.py
smereu: @RobPasMue Thank you for the clarification! Removed

tests/integration/test_prepare_tools.py
smereu: Changed to using names @RobPasMue  In case you wonder why there is a new file, the one I was using was a CAD file that just so happened to have 3 bodies with the same :-).  So I imported in SC to change the name and I am now using the SC file

RobPasMue (approved):
Implementation looks good @smereu - thanks for handling all my comments!

PR #1904, opened 2025-04-11
build: bump grpcio dependencies

RobPasMue:
Closes  #1655

PR #1913, opened 2025-04-15
build: bump grpcio limits and handle erratic gRPC channel creation

### LINKED ISSUE #1655: Upgrade upper limits on grpcio and grpcio-health-checking
Getting recurrent issues on versions 1.69.X and above. Had to ignore their bumps in https://github.com/ansys/pyansys-geometry/pull/1618 and https://github.com/ansys/pyansys-geometry/pull/1654 temporarily. Eventually we should move the upper limit to <2.X

RobPasMue:
Closes #1655


RobPasMue:
Follow up links - see https://github.com/ansys/pyansys-geometry/pull/1904#issuecomment-2797032223

PR #1922, opened 2025-04-18
Feat/update with delta

umutsoysalansys:
## Description

## Issue linked

## Checklist


src/ansys/geometry/core/tools/problem_areas.py
```diff
@@ -237,15 +303,17 @@ def fix(self) -> RepairToolMessage:
             return RepairToolMessage(False, [], [])
 
         parent_design = get_design_from_edge(self.edges[0])
+
         response = self._repair_stub.FixInexactEdges(
             FixInexactEdgesRequest(inexact_edge_problem_area_id=self._grpc_id)
         )
-        parent_design._update_design_inplace()
+
```
RobPasMue, line 310: After the fix operation... we are not updating the design? This seems problematic

src/ansys/geometry/core/designer/design.py
```diff
@@ -1220,3 +1220,139 @@ def _update_design_inplace(self) -> None:
 
         # Read the existing design
         self.__read_existing_design()
+
+    def update_from_tracker(self, tracker_response):
```
RobPasMue, line 1224: I would name this method as protected (i.e. ``_update_from_tracker``). Ideally users shouldn't have to call it

RobPasMue, line 1224: Missing typehints in function definition

src/ansys/geometry/core/designer/design.py
```diff
@@ -1220,3 +1220,139 @@ def _update_design_inplace(self) -> None:
...
+        """
+        Update the design with the changed bodies while preserving unchanged ones.
+
+        Parameters
+        ----------
+        changed_bodies : list[dict]
+            A list of dictionaries representing changed body information. Each dictionary
+            should contain 'id', 'name', and 'is_surface' keys.
+        """
```
RobPasMue, line 1233: I have the feeling that the parameter name changed to "tracker_response" instead of "changed_bodies"

src/ansys/geometry/core/tools/problem_areas.py
```diff
@@ -131,11 +178,19 @@ def fix(self) -> RepairToolMessage:
         response = self._repair_stub.FixDuplicateFaces(
             FixDuplicateFacesRequest(duplicate_face_problem_area_id=self._grpc_id)
         )
```
RobPasMue, line 180: Oh I just realized about this.. we need to move this into the gRPC folder implementation instead of having at this level raw gRPC requests. That would avoid method duplication as well for serialization.

Let's do this in a follow up PR I think

doc/changelog.d/1808.fixed.md
RobPasMue, line 1: I think this has to be deleted -- comes from an old PR that already got merged.

doc/changelog.d/1806.added.md
RobPasMue, line 1: I think this has to be deleted -- comes from an old PR that already got merged.

src/ansys/geometry/core/_grpc/_services/v0/repair_tools.py
```diff
@@ -434,3 +455,91 @@ def serialize_issue(issue):
...
+
+        def serialize_entity_identifier(entity):
+            """Serialize an EntityIdentifier object into a dictionary."""
+            return {
+                "id": entity.id,
+                "type": entity.type,
+            }
+
+    @staticmethod
+    def serialize_tracker_command_response(response) -> dict:
```
RobPasMue, line 497: Method shadowing is happening -- only the second implementation will be used. Why do we have it duplicated? Can we get rid of the one that is not used?

Also.. can we avoid making it a "staticmethod" and make it a normal, protected method? (i.e. ``_serialize_tracker_command_response``)

src/ansys/geometry/core/designer/design.py
```diff
@@ -1220,3 +1220,139 @@ def _update_design_inplace(self) -> None:
...
+        print("Updating design using the tracker...")
```
RobPasMue, line 1234: Use the logger rather than print statements

src/ansys/geometry/core/designer/design.py
```diff
@@ -1220,3 +1220,139 @@ def _update_design_inplace(self) -> None:
...
+
+        # Function to update a body if it exists
+        def update_body(existing_body, body_info):
+            existing_body.name = body_info["name"]
+            existing_body._template._is_surface = body_info.get("is_surface", False)
+            print(f"Updated body: {body_info['name']} (ID: {body_info['id']})")
```
RobPasMue, line 1240: Same here in regards to the print statement

src/ansys/geometry/core/designer/design.py
```diff
@@ -1220,3 +1220,139 @@ def _update_design_inplace(self) -> None:
...
+        # Loop through all deleted bodies from the tracker
+        for body_info in tracker_response["deleted_bodies"]:
+            body_id = body_info["id"]
+            removed = False  # Track if a body was removed
+
+            # First, check bodies at the root level
+            for body in self.bodies:
+                if body.id == body_id:
+                    body.is_alive = False
+                    # self.bodies.remove(body)
```
RobPasMue, line 1325: I think we have to activate this -- self.bodies should only contain "alive" bodies. However, the proper approach is to mark the body as "dead" by changing the bool and removing it from the list. Did you experience problems when doing this?

src/ansys/geometry/core/tools/problem_areas.py
```diff
@@ -78,10 +77,58 @@ def id(self) -> str:
         """The id of the problem area."""
         return self._id
 
-    @abstractmethod
-    def fix(self):
-        """Fix problem area."""
-        raise NotImplementedError("Fix method is not implemented in the base class.")
+    @staticmethod
```
RobPasMue, line 80: Was this intended? I think the abstractmethod for "fix" is still needed.

RobPasMue (commented):
Looks like all the tests have gone through in the end @umutsoysalansys! Are you still experiencing issues locally? I left a first review with some comments. Let me know once I can do a final review. Thanks for all the work here, this looks promising!

src/ansys/geometry/core/tools/repair_tools.py
```diff
@@ -479,14 +480,20 @@ def find_and_fix_short_edges(
...
+            parent_design.update_from_tracker(response["complete_command_response"])
+
         message = RepairToolMessage(
             success=response["success"],
             found=response["found"],
             repaired=response["repaired"],
             created_bodies=[],
             modified_bodies=[],
         )
+        parent_design.update_from_tracker(response["complete_command_response"])
```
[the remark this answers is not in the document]
umutsoysalansys: You are right, thanks for catching.

tests/integration/test_repair_tools.py
```diff
@@ -285,17 +285,6 @@ def test_fix_interference(modeler: Modeler):
...
-    design = modeler.open_file(FILES_DIR / "DuplicateFaces.scdocx")
-    assert len(design.bodies) == 7
-    areas = modeler.repair_tools.find_duplicate_faces(design.bodies)
-    assert len(areas) == 6
-    for area in areas:
-        area.fix()
-    assert len(design.bodies) == 1
-
-
 def test_find_and_fix_extra_edges_problem_areas(modeler: Modeler):
```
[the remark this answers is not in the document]
umutsoysalansys: yeah sure, only diff in this file is marked with todo.

src/ansys/geometry/core/designer/design.py
umutsoysalansys: makes sense, done.

src/ansys/geometry/core/designer/design.py
RobPasMue: I think the logs would be nice to have nonetheless. Maybe we can simplify them.

And yes, use self._grpc_client.log.** for performing logging, whatever log level message you want to put. This is client-side logging.

Some might be good at debug, others might be good at info. Thanks @umutsoysalansys!

src/ansys/geometry/core/designer/design.py
umutsoysalansys: is_alive seems like only a getter not setter? do we want to add a setter for it too? If we are removing it from the list, I am not sure what does it do to mark it as dead?

RobPasMue: Ah yes -- but as a library developer you can make use of the ``_is_alive`` protected attribute to put it to False :) users aren't expected to make use of it, but we need to use it in circumstances like this one for example.

If _is_alive is False, it won't allow a user to perform gRPC calls on that object. And avoiding kernel crashes on the server side.

So, as a summary: don't create a setter, just use the protected attribute (_is_alive) and set it to False. And then, if a user is holding a reference to that object, they will also get the update and will no longer be allowed to perform server calls on that object =)

umutsoysalansys:
> Looks like all the tests have gone through in the end @umutsoysalansys! Are you still experiencing issues locally? I left a first review with some comments. Let me know once I can do a final review. Thanks for all the work here, this looks promising!

Yes because it is also behind the feature flag which is set to False as default ( so update design inplace is activated) The full test suit with the tracker is still failing in 2 tests because of some potential leak on Geometry Service side but all tests are passing individually and still following up on that issue with @david-gorman @b-matteo. Once it is resumed we can working on changing the flag with extensive testing.

smereu:
@RobPasMue Thanks for the comment. If the issue is easily reproduced, we can try to look at it this week as it appears to be entirely on the server side. @david-gorman please correct if me I am wrong but you were not able to reproduce it your machine. My guess is that there is some memory leak that grows the memory and David's machine being a powerful one with a lot of memory may be able to survive....

src/ansys/geometry/core/__init__.py
```diff
@@ -70,3 +70,6 @@
 
 DOCUMENTATION_BUILD: bool = os.environ.get("PYANSYS_GEOMETRY_DOC_BUILD", "false").lower() == "true"
 """Global flag for the documentation to use the proper PyVista Jupyter backend."""
+
+USE_TRACKER_TO_UPDATE_DESIGN: bool = True
```
RobPasMue, line 74:
```suggestion
USE_TRACKER_TO_UPDATE_DESIGN: bool = False
```
Reverting to ``False`` as discussed internally - I'm glad it's working!!! Thanks @umutsoysalansys

RobPasMue (approved):
LGTM! Thanks @umutsoysalansys!

src/ansys/geometry/core/tools/repair_tools.py
```diff
@@ -529,6 +532,20 @@ def find_and_fix_short_edges(
...
+
+        message = RepairToolMessage(
+            success=response["success"],
+            found=response["found"],
+            repaired=response["repaired"],
+            created_bodies=[],
+            modified_bodies=[],
+        )
+        return message
         parent_design._update_design_inplace()
```
RobPasMue, line 550: There seems to be some unreachable code here @umutsoysalansys - I think this part of the code needs some attention. Could you review it?

RobPasMue (changes requested):
Blocking until unreachable code issue is resolved.

src/ansys/geometry/core/tools/repair_tools.py
umutsoysalansys: cleaned, the merge artifacts.

src/ansys/geometry/core/__init__.py
RobPasMue, line 74:
```suggestion
USE_TRACKER_TO_UPDATE_DESIGN: bool = False
```

PR #2048, opened 2025-06-18
Feat: allow logos Linux 26 1

RyanJWard:
## Description
With the latest 26.1 Linux service prepare logo tools should be fixed

## Checklist

### DIFF
--- a/src/ansys/geometry/core/tools/prepare_tools.py
+++ b/src/ansys/geometry/core/tools/prepare_tools.py
@@ -300,8 +300,8 @@ class PrepareTools:
         """
         from ansys.geometry.core.designer.body import Body
 
-        if BackendType.is_linux_service(self._grpc_client.backend_type):
-            # not yet available in Linux
+        if BackendType.is_linux_service(self._grpc_client.backend_type) and self._grpc_client.backend_version < (26, 1, 0):
+            # not yet available on Linux until 26.1.0
             LOG.warning("Logo detection not available on Linux")
             return
 
@@ -346,8 +346,8 @@ class PrepareTools:
         """
         from ansys.geometry.core.designer.body import Body
 
-        if BackendType.is_linux_service(self._grpc_client.backend_type):
-            # not yet available in Linux
+        if BackendType.is_linux_service(self._grpc_client.backend_type) and self._grpc_client.backend_version < (26, 1, 0):
+            # not yet available on Linux until 26.1.0
             LOG.warning("Logo detection not available on Linux")
             return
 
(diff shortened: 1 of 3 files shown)


tests/test_connection.py
```diff
@@ -360,3 +366,54 @@ def test_api_versions_reader():
...
+    ):
+        prepare_and_start_backend(
+            backend_type=BackendType.WINDOWS_SERVICE, version=1900, product_version=1901
+        )
+
+
+@pytest.mark.skipif(
+    sys.platform.startswith("linux"),
+    reason="Test skipped on Linux because it is specific to Windows backends.",
+)
```
RobPasMue, line 395: We should use the is_linux_service() call here rather than using "sys". You could be testing against a remote Linux service from a Windows environment.. and then the test would get executed :)

RobPasMue: Oh... now that I think of it. You might actually have to go with "sys". You are attempting to launch the backend so you can't perform that call... really sorry!

RobPasMue: Or you might bot even need to skip it, right? Because the execution will never happen so you are safe in any case

RobPasMue: Just checked, makes sense. Those errors are only happening if running on a Windows system. I would skip the test if anybody tries to run it on anything else than Windows... so:

```py
pytest.mark.skipif(
    os.name != "nt",
    reason="Test skipped on Linux because it is specific to Windows backends.",
)
```

RobPasMue: Sorry for all the fuss @RyanJWard ... I didn't read the test properly the first time

jonahrb (approved):

src/ansys/geometry/core/tools/prepare_tools.py
```diff
@@ -372,10 +372,8 @@ def find_and_remove_logos(
         """
         from ansys.geometry.core.designer.body import Body
 
-        if BackendType.is_linux_service(
-            self._grpc_client.backend_type
-        ) and self._grpc_client.backend_version < (26, 1, 0):
-            # not yet available on Linux until 26.1.0
+        if BackendType.is_linux_service(self._grpc_client.backend_type):
+            # not yet available in Linux
```
RobPasMue: Oh so it's not available - generally-speaking - on Linux? I thought it would have been available in 26R1

RobPasMue (commented):
Thanks for this fix!

tests/integration/test_prepare_tools.py
```diff
@@ -126,19 +126,19 @@ def test_enhanced_share_topology(modeler: Modeler):
...
+    # no logos should be found if max height is not given
     assert len(result.face_ids) == 0
     result = modeler.prepare_tools.find_logos(max_height=0.005)
     assert len(result.face_ids) == 147
     success = modeler.prepare_tools.find_and_remove_logos(max_height=0.005)
+    # Skip the rest of the test if running on a Linux service backend
+    if BackendType.is_linux_service(modeler.client.backend_type):
+        return
     assert success is True
     assert len(body.faces) == 42
```
RobPasMue, line 143: Why had the rest to be skipped?

PR #2169, opened 2025-08-12
feat: option to write body facets on save

jonahrb:
## Description
Write body facets into the scdocx on save

## Checklist

### DIFF
--- a/src/ansys/geometry/core/designer/design.py
+++ b/src/ansys/geometry/core/designer/design.py
@@ -251,6 +251,7 @@ class Design(Component):
         self,
         file_location: Path | str,
         format: DesignFileFormat = DesignFileFormat.SCDOCX,
+        write_body_facets: bool = False,
     ) -> None:
         """Export and download the design from the server.
 
@@ -275,7 +276,9 @@ class Design(Component):
         if self._modeler.client.backend_version < (25, 2, 0):
             received_bytes = self.__export_and_download_legacy(format=format)
         else:
-            received_bytes = self.__export_and_download(format=format)
+            received_bytes = self.__export_and_download(
+                format=format, write_body_facets=write_body_facets
+            )
 
         # Write to file
         file_location.write_bytes(received_bytes)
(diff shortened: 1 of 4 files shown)


RyanJWard (changes requested):
@jonahrb We need to skip these tests for older version and we might need a minimum version right?

src/ansys/geometry/core/designer/design.py
```diff
@@ -351,14 +364,18 @@ def __export_and_download(self, format: DesignFileFormat) -> bytes:
...
+                )
             except Exception:
                 self._grpc_client.log.warning(
                     f"Failed to download the file in {format} format."
                     " Attempting to stream download."
                 )
                 # Attempt to download the file via streaming
-                response = self._grpc_client.services.designs.stream_download_export(format=format)
+                response = self._grpc_client.services.designs.stream_download_export(
+                    format=format, write_body_facets=write_body_facets
```
RobPasMue, line 377: Agreed.. the write_body_facets, in case it is True, should be verified against the backend version. If True and backend_version < 26.1 then we need to throw a warning and inform that this option will not be considered since the feature wasn't available prior to 26.1

RobPasMue: This comment applies to the three methods. Maybe you want to come up with a common method and reuse it. Something like ``check_write_body_facets_input(write_body_facets, backend_version)``.

RobPasMue: I would do something like this in fact:

```py
        # HACK: we should inform the user that NURBS curve sketches are not supported
        # prior to 26R1... and if passed, raise an error
        _nurbs_curves_compatibility(kwargs["backend_version"], request.geometries)
 ```
 
 This is already being done inside the ``_grpc/_services/v0/bodies.py`` file. Could you implement this @jonahrb ?

RyanJWard (approved):

RobPasMue (approved):
Awesome, LGTM!

PR #2226, opened 2025-09-12
feat: for 261 dotnet 8 is included in core service

RyanJWard:
## Description
We made build changes for 261 to include dotnet 8

## Checklist


b-matteo (approved):
LGTM

src/ansys/geometry/core/connection/product_instance.py
```diff
@@ -464,29 +464,41 @@ def prepare_and_start_backend(
             # Verify dotnet is installed
             import shutil
 
-            if not shutil.which("dotnet"):
-                raise RuntimeError(
-                    "Cannot find 'dotnet' command. "
-                    "Please install 'dotnet' to use the Ansys Geometry Core Service. "
-                    "At least dotnet 8.0 is required."
+            if version < 261:  # Only execute this block if version is less than 261
```
RobPasMue, line 467: We might have to be more thoughtful than just using the version. This part is specifically for Linux and, in most cases, users do not have AWP_ROOTXXX defined... which means that they are going into this through defining the ``ANSYS_GEOMETRY_SERVICE_ROOT``

I would do as follows:

1- Search if the binary exists - if it exists, use it. (this will only be true already above 261)
2- If the binary doesn't exist then we attempt to use dotnet to run the DLL (this will also address the below 261 request)

src/ansys/geometry/core/connection/product_instance.py
```diff
@@ -465,29 +465,45 @@ def prepare_and_start_backend(
...
-            # At least dotnet 8.0 is required
-            # private method and controlled input by library - excluding bandit check.
-            if subprocess.check_output(["dotnet", "--version"]).decode("utf-8").split(".")[0] < "8":  # nosec B607, B603
-                raise RuntimeError(
-                    "Ansys Geometry Core Service requires at least dotnet 8.0. "
-                    "Please install a compatible version."
+            if any(
+                CORE_GEOMETRY_SERVICE_EXE.replace(".exe", "") in file.name
+                for file in Path(root_service_folder).iterdir()
+            ):
```
RobPasMue, line 499: Shouldn't this be an else directly? Since the first if is the exact contrary.

RyanJWard: @RobPasMue It is, I wasn't sure if we wanted to explicitly check that the file is there before trying to launch it. I can change this to an else

PR #2239, opened 2025-09-19
fix: beam issues in 25R2 and above

RobPasMue:
Reported in #2238


RobPasMue:
Fixed in https://github.com/ansys/pyansys-geometry/commit/afb33f7bdae777732ccb90273a21461c2f80043c by @jacobrkerstetter !

Issue #2543, opened 2026-01-30
Tracker fails in test_component_make_independent


jacobrkerstetter:
When running test_component_make_independent with tracker enabled, the test fails and updates all bodies occurrences.

PR #2592, opened 2026-02-18
fix: clean use of ids

smereu:
## Description
Issue#1 Some parts of pygeometry code were written using guids for geometry entities. While it seemed harmless, it prevents matching of ids when the proper id is recovered and used for comparison (e.g. for tracking).
The only exception is the design which doesn't have an id
The branch also corrects one issue in creating occurrences via translation

Issue #2 The call to component make independent make very significant changes the hierarchy. There was no call to update_design_in_place nor tracker so any code executed after this call would have been potentially accessing invalid variables. In the issue tracked here a follow-up call (extrude) was trying to update (in place or via tracker). Since the tracker is atomic nothing was coming back, but the call to update_design_in_place was restoring ability to access correct data. I tried to put the call from tracker but ran into issues. 1) Minor - the server response was not right 2) The update from tracker returns that pretty much everything has changed and we have some issues. I will try to sort them out but for the time being I put a call to update_design_in_place. If the changes are major the savings using tracker are likely minor anyway,

## Issue linked
 Closes https://github.com/ansys/pyansys-geometry/issues/2064

Related to https://github.com/ansys/pyansys-geometry/issues/2543

## Checklist

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -268,8 +268,7 @@ def fake_modeler_old_backend_251(modeler: Modeler):
 @pytest.fixture(scope="function")
 def fake_modeler_old_backend_252(modeler: Modeler):
     currentbackend = modeler._grpc_client._backend_version
-    currentservices = modeler._grpc_client._services\
-
+    currentservices = modeler._grpc_client._services
     # Check if server supports v0 protocol
     if not GeometryApiProtos.V0.verify_supported(modeler._grpc_client.channel):
         pytest.skip("Server does not support v0 protocol needed for this test")


smereu:
I need to still run all the tests as I am having issue with geometry core service but wanted to publish the PR in case there are comments

src/ansys/geometry/core/_grpc/_services/v0/components.py
```diff
@@ -97,7 +99,7 @@ def set_placement(self, **kwargs) -> dict:  # noqa: D102
 
         # Create the direction and point objects
         translation = (
-            from_unit_vector_to_grpc_direction(kwargs["translation"].normalize())
+            from_unit_vector_to_grpc_direction(kwargs["translation"])
```
RobPasMue: Nice catch @smereu !! Definitely, the translation vector shouldn't be normalized. Otherwise it defeats its purpose - we are not simply identifying a direction

src/ansys/geometry/core/designer/component.py
```diff
@@ -257,9 +259,14 @@ def __init__(
...
             # This is an independent Component - Create new Part and MasterComponent
-            p = Part(uuid.uuid4() if not self._template else self._template, f"p_{name}", [], [])
+            p = Part(
+                uuid.uuid4() if not self._component else self._component_part_master_id,
+                f"p_{name}",
+                [],
+                [],
+            )
             master = MasterComponent(
-                uuid.uuid4() if not self._template else self._component.master_id,
```
RobPasMue: I'm curious - why don't we use ``_template`` anymore?

RobPasMue (commented):
Looking really good @smereu !

PR #2618, opened 2026-03-02
ci: skip system-level quarto installation

RobPasMue:
As title says - we already have quarto-cli as a package dependency

### DIFF
--- a/doc/changelog.d/2544.test.md
+++ b/doc/changelog.d/2544.test.md
@@ -1,0 +1,1 @@
+Enable disabled tests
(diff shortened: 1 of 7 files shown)


pyproject.toml
```diff
@@ -71,7 +71,8 @@ dev = [
   { include-group = "tests" },
 ]
 doc = [
-  "ansys-sphinx-theme[autoapi]==1.7.0",
+  # use ansys-sphinx-theme main branch to get the latest bug fixes and features, see
```
RobPasMue:
```suggestion
```

PR #2625, opened 2026-03-04
feat: allow deployment of custom images

RobPasMue:
## Description
Coming from @ans-spohekar's request - being able to pass in a custom image name/tag to deploy

```py
from ansys.geometry.core import launch_docker_modeler

# Using a custom Docker image
modeler = launch_docker_modeler(
    image="myregistry.com/my-geometry:custom-tag"
)
```

Make sure to pass the full image name though, not just the tag. Give it a try to this branch locally on your side if possible

### DIFF
--- a/doc/changelog.d/2624.documentation.md
+++ b/doc/changelog.d/2624.documentation.md
@@ -1,0 +1,1 @@
+Add warning for stale variables on design class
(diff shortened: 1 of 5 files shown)


RobPasMue:
Let's not merge this one until it has been validated by @ans-spohekar's side

PR #2773, opened 2026-05-14
feat: improved autocomplete experience

RobPasMue:
## Description
Autocomplete on IDEs was becoming buggy because of the way the typehints were passed in the decorator definitions. This fixes this issue.

## Issue linked
None

## Checklist

### DIFF
--- a/src/ansys/geometry/core/misc/checks.py
+++ b/src/ansys/geometry/core/misc/checks.py
@@ -23,10 +23,11 @@
 
 from collections.abc import Callable, Iterable
 import functools
-from typing import TYPE_CHECKING, Any, Type, TypeVar, get_type_hints
+from typing import TYPE_CHECKING, Any, Type, TypeVar
 import warnings
 
 from beartype import beartype as _beartype
+from beartype.roar import BeartypeCallHintForwardRefException as _BeartypeForwardRefError
 import numpy as np
 from pint import Unit
 import semver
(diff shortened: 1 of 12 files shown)


src/ansys/geometry/core/misc/checks.py
```diff
@@ -80,12 +84,12 @@ def get_design_ref(obj) -> "Design":
...
         # Finally, call method
-        return method(self, *args, **kwargs)
+        return method(*args, **kwargs)
 
-    return wrapper
+    return wrapper  # type: ignore[return-value]
 
 
-def check_input_types(func):
+def check_input_types(func: _F) -> _F:
```
MaxJPRey, line 92: If I am correct, now we catches any exception while building the beartype wrapper
and permanently disables beartype for that function if a forward-ref problem occurs at call time.
Which makes for autocomplete / typing friendliness. Nice!

Some type-checking failures could get silently bypassed, but I think it is aceptable in this project to trade some runtime type-check strictness for better editor/typehint behavior.

MaxJPRey (approved):
LGTM.

src/ansys/geometry/core/misc/checks.py
RobPasMue: Precisely. It is all caused because of forward references really.

If any forward reference occurs and Beartype is incapable of resolving it, the decorator will fail anyway (it wasn't working previously already). In fact, in those cases, we typically leveraged the ``check_*`` functions in this module. We also implemented and general on-off switch in #2767 that not only disables the beartype check but also the check functions to get enhanced performance in production.

The autocomplete friendliness is caused by the shadowing effect of decorators. We need to pass the function typehints properly, otherwise IDEs completely ignore their signatures..

PR #2797, opened 2026-05-21
fix: improve named selection performance

jacobrkerstetter:
## Description
adding small verified check, improves operations that call all properties from the ns by only pinging server once

## Checklist

### DIFF
--- a/src/ansys/geometry/core/designer/selection.py
+++ b/src/ansys/geometry/core/designer/selection.py
@@ -307,34 +307,38 @@ class NamedSelection:
         self.__verify_ns()
         self._verified = True
 
-        # Convert None to empty lists
-        bodies = bodies if bodies is not None else []
-        faces = faces if faces is not None else []
-        edges = edges if edges is not None else []
-        beams = beams if beams is not None else []
-        design_points = design_points if design_points is not None else []
-        components = components if components is not None else []
-        vertices = vertices if vertices is not None else []
+        try:
+            # Convert None to empty lists
+            bodies = bodies if bodies is not None else []
+            faces = faces if faces is not None else []
+            edges = edges if edges is not None else []
+            beams = beams if beams is not None else []
+            design_points = design_points if design_points is not None else []
+            components = components if components is not None else []
+            vertices = vertices if vertices is not None else []
 
-        new_ns = NamedSelection(
-            self._name,
-            self._design,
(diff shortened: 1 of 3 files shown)


src/ansys/geometry/core/designer/selection.py
```diff
@@ -97,6 +97,7 @@ def __init__(
         self._name = name
         self._design = design
         self._grpc_client = grpc_client
+        self._verified = False
```
RobPasMue, line 100: I'm worried about this one.. how do we know if somebody modified the NS without using this object? I'm thinking about geometry command operations for example, or body deletions from a component.. they will not get reflected here automatically.

RobPasMue (changes requested):
TBDiscussed

src/ansys/geometry/core/designer/selection.py
jacobrkerstetter: Would double underscores work in that case? I only use the verified flag on operations within the class that we control to reduce the amount of server calls to 1. I definitely don't intend on it being used elsewhere.

RobPasMue: I might not have been super clear with my first comment

The single underscore is good enough for saying it is for internal usage only, no need to worry about it.

I'm concerned by the fact that the NS might go out of sync in a heartbeat... and we won't have a way to easily reflect those changes anymore. Am I missing something?

jacobrkerstetter: Definitely understand what you are saying, but I think the misunderstanding is that this is only used in methods like __repr__ or add/remove so that the calls to self.__verify_ns() are skipped in the individual properties. The verified flag is set to False at the end of the method. 

Since we have this control on guaranteeing nothing else can modify the NS during the time it is True, it should be safe. Then, at the end of these "safe" methods, we return it to False, and any changes will be reflected since we are forcing the self.__verify_ns() calls to be made again.

RobPasMue (approved):
Ah perfect! I missed the "finally" behavior!! Thanks @jacobrkerstetter

src/ansys/geometry/core/designer/selection.py
RobPasMue: All good - I missed the finally behavior, that one is on me... thanks @jacobrkerstetter !

PR #2821, opened 2026-06-02
fix: add process isolation to startup

### LINKED ISSUE #2820: Geometry service currently started without process isolation is causing issues for pytest
https://github.com/ansys/pyansys-geometry/blob/b8ac8028bfe05ee5212994e9aa3c790405ef47b7/src/ansys/geometry/core/connection/product_instance.py#L721-L728

plus

https://github.com/ansys/pyansys-geometry/blob/b8ac8028bfe05ee5212994e9aa3c790405ef47b7/src/ansys/geometry/core/connection/product_instance.py#L168-L179

These are causing issues for us when testing on Windows

On Microsoft Windows, CTRL_BREAK_EVENT always goes to the whole group, and it seems there's no way to scope it to a single PID (https://learn.microsoft.com/en-us/windows/console/console-process-groups)

Should `subprocess.Popen` call have something like a `creationflags=CREATE_NEW_PROCESS_GROUP` argument also, so that the process is isolated? Or would that not work with the current approach you guys have here?

Otherwise when testing with `pytest` we run into issues in packages that depend on this one (private repo link with more details should show up below in this issue tracker, internal developers have access)

Please let us know if you have any other workarounds

jacobrkerstetter:
## Description
- create new process group on Popen so that the gracefully close doesn't kill all processes

## Issue linked
Closes #2820 

## Checklist

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -168,7 +168,8 @@ jobs:
           fi
 
   testing-windows:
-    name: Testing and coverage (Windows) (${{ matrix.tracker-label }}, ${{ matrix.docker-image }})
+    name: Testing and coverage (Windows) (${{ matrix.proto-version }}, ${{ matrix.tracker-label }}, ${{ matrix.docker-image
+      }})
     needs: [smoke-tests, manifests]
     runs-on: windows-latest
     continue-on-error: ${{ matrix.experimental }}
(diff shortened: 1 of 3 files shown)


raph-luc (dismissed):
This fixes the issue described in https://github.com/ansys/pyansys-geometry/issues/2820 on my end

raph-luc (dismissed):

RobPasMue (approved):

PR #2847, opened 2026-06-10
test: Increased test coverage 

Gonzi09:
Expanded coverage for selection.py, added missing coverage for design and added missing docstrings 

## Checklist

### DIFF
--- a/tests/integration/test_design.py
+++ b/tests/integration/test_design.py
@@ -5531,25 +5531,6 @@ def test_update_body_properties(modeler: Modeler):
     assert body._template._is_surface is True
 
 
-def test_clear_body_cache_for_part(modeler: Modeler):
-    """Test _clear_body_cache_for_part clears caches for all matching components."""
-    design = modeler.create_design("clear_body_cache_test")
-
-    comp = design.add_component("TestComp")
-    part = comp._master_component.part
-
-    sketch = Sketch()
-    sketch.box(Point2D([0, 0]), 1, 1)
-    comp.extrude_sketch("TestBody", sketch, 1)
-
-    _ = comp.bodies
-    assert "bodies" in comp.__dict__
-
-    design._clear_body_cache_for_part(part)
-
-    assert "bodies" not in comp.__dict__
-
-
 def test_clear_body_cache_for_part_nested(modeler: Modeler):
     """Test _clear_body_cache_for_part clears caches in nested components."""
     design = modeler.create_design("clear_cache_nested")
--- a/tests/integration/test_design.py
+++ b/tests/integration/test_design.py
@@ -369,10 +369,7 @@ def test_update_from_tracker_created_parts(modeler: Modeler):
 
 
 def test_update_from_tracker_deleted_parts_with_fallback(modeler: Modeler):
-    """Test _update_from_tracker() for deleted parts.
-
-    Verifies parts are marked inactive and warnings are logged for missing parts.
-    """
+    """Test _update_from_tracker() marks deleted parts as inactive and logs warnings for missing parts."""
     design = modeler.create_design("update_tracker_deleted")
 
     mock_part = Mock()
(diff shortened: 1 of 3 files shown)


tests/integration/test_design.py
```diff
@@ -6078,6 +6747,219 @@ def test_tracking_captures_boolean_operations(modeler: Modeler):
     assert len(unite_changes["deleted_bodies"]) == 1
 
 
+def _tracker_payload(**overrides):
```
RobPasMue: This method is duplicate I believe.. can we clean up the PR?

RobPasMue (dismissed):

tests/integration/test_design.py
```diff
@@ -4907,6 +4942,624 @@ def test_get_edge_bounding_box(modeler: Modeler):
...
+        return payload
+
+    return _factory
+
+
+@pytest.fixture
+def unit_box_sketch() -> Sketch:
+    sketch = Sketch()
+    sketch.box(Point2D([0, 0]), 1, 1)
+    return sketch
```
jacobrkerstetter: Let's move the fixtures to the top or to the conftest file if they can be reused. Then you can import them to this file.

tests/integration/test_design.py
```diff
@@ -4907,6 +4942,624 @@ def test_get_edge_bounding_box(modeler: Modeler):
...
+            "modified_components": [],
+            "deleted_components": [],
+            "created_bodies": [],
+            "modified_bodies": [],
+            "deleted_bodies": [],
+        }
+        payload.update(overrides)
+        return payload
+
+    return _factory
```
jacobrkerstetter: I'm seeing the duplication now... here and where Roberto commented

tests/integration/test_design.py
```diff
@@ -6124,3 +6777,217 @@ def test_move_bodies_to_component(modeler: Modeler):
...
+def test_find_and_remove_part_sets_alive_flag(modeler: Modeler):
+    """Test _find_and_remove_part marks part as not alive."""
+    design = modeler.create_design("cov_remove_part_flag")
+    part = design._master_component.part
+    part._is_alive = True
+
+    removed = design._find_and_remove_part({"id": part.id})
+
+    assert removed is True
+    assert part._is_alive is False
```
jacobrkerstetter, line 6746: See if there's some places we can combine tests. For example, the only coverage we're adding here is testing the part._is_alive flag in this case. This could be done in any test that also has a step where a part in removed. I think we can reduce overall lines of code like this/reduce setup + teardown from having so many tests

tests/integration/test_design.py
```diff
@@ -6124,3 +6777,217 @@ def test_move_bodies_to_component(modeler: Modeler):
...
+    updated = design._find_and_update_component(
+        {"id": nested.id, "name": "Updated"}, design.components
+    )
+    not_found = design._find_and_update_component(
+        {"id": "missing", "name": "Nope"}, design.components
+    )
+
+    assert updated is True
+    assert nested.name == "Updated"
+    assert not_found is False
```
jacobrkerstetter, line 6811: This would be an example of two (or maybe more below?) tests that can be combined without a lot of effort. Use the first set up to test both cases

jacobrkerstetter (changes requested):
Couple changes but overall looks good. Lots of coverage added! 😄

tests/integration/test_design.py
```diff
@@ -6079,6 +6680,220 @@ def test_tracking_captures_boolean_operations(modeler: Modeler):
...
+        "deleted_parts": [],
+        "created_components": [],
+        "modified_components": [],
+        "deleted_components": [],
+        "created_bodies": [],
+        "modified_bodies": [],
+        "deleted_bodies": [],
+    }
+    payload.update(overrides)
+    return payload
```
jacobrkerstetter: This can be removed from this file now

jacobrkerstetter (changes requested):
Still has the one duplicate _tracker_payload left. Maybe after this is merged we can consider moving some tests to a "test_named_selections.py" and "test_tracker.py" - just a thought!

jacobrkerstetter (approved):
Thanks for those changes! Looks very clean now

PR #2872, opened 2026-06-18
feat: switch to uv

RobPasMue:
## Description
Switching build system to uv - adapting CI/CD accordingly and documentation

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -59,7 +59,7 @@ jobs:
     name: Actions Security
     runs-on: ubuntu-latest
     steps:
-      - uses: ansys/actions/check-actions-security@d946b24b9a765f4169bcc94afdb27bd1a0533741 # v10.3.2
+      - uses: ansys/actions/check-actions-security@main
         with:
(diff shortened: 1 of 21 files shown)


.github/workflows/ci_cd.yml
```diff
@@ -59,7 +59,7 @@ jobs:
     name: Actions Security
     runs-on: ubuntu-latest
     steps:
-      - uses: ansys/actions/check-actions-security@d946b24b9a765f4169bcc94afdb27bd1a0533741 # v10.3.2
+      - uses: ansys/actions/check-actions-security@main
```
jacobrkerstetter, line 62: Is there a reason we aren't using a particular release of the security actions anymore?

jacobrkerstetter (approved):

.github/workflows/ci_cd.yml
RobPasMue: Mostly because the version of zizmor pinned down inside it is pretty old. Based on new zizmor releases, the security measures we are taking are valid. In the old versions, this was not the case. We should be able to revert in the next ansys/actions release

PR #2888, opened 2026-06-25
feat: search for component by name

jacobrkerstetter:
## Description
As title says. Provides new recursive search + test

## Checklist


src/ansys/geometry/core/designer/component.py
```diff
@@ -1675,6 +1675,38 @@ def search_component(self, id: str) -> Union["Component", None]:
         # If you reached this point... this means that no component was found!
         return None
 
+    @check_input_types
+    def search_component_by_name(self, name: str) -> Union["Component", None]:
```
RobPasMue: Is it possible to have several components with the same name? In such case we would only be returning the first hit I guess.. right?

RobPasMue (commented):
I like the idea - I'm just worried about the multiple hit scenario. But that would imply that we need to traverse the entire design no matter what... even if we found a hit. What do you think @jacobrkerstetter ?

src/ansys/geometry/core/designer/component.py
```diff
@@ -1675,6 +1675,45 @@ def search_component(self, id: str) -> Union["Component", None]:
         # If you reached this point... this means that no component was found!
         return None
 
+    @check_input_types
+    def search_component_by_name(self, name: str) -> Union[list["Component"], "Component", None]:
```
RobPasMue: Three return types might be difficult to handle. It's preferable to return ``list["Component"]`` in this case. That way users can handle things consistently --> See if the list has elements, retrieve the first element, use all of them etc.

jacobrkerstetter: Would it be a breaking change to get rid of the `"Component"` return type in favor of `list["Component"]`? Is that ok to do here?

RobPasMue: It can't be a breaking change since this API didn't exist before =) it would be different if we changed the ``search_component`` API but that one expects a specific ID.. and returns either None or a specific Component

RobPasMue (approved):

PR #2922, opened 2026-07-07
feat: add selection builder with body api's

jacobrkerstetter:
## Description
Adds the SelectionBuilder APIs to make selecting objects by criteria easier.
This PR uses BodySelection as the initial PoC to show the initial capabilities.

## Checklist


.github/agents/method-implementation-agent.agent.md
```diff
@@ -76,4 +76,7 @@ Once the implementation is complete, pass the task to the writing-tests agent us
 ## Virtual environment
 Do not search for a `.venv` directory. Instead, read the `PYANSYS_VENV` environment variable to obtain the path to the virtual environment, then activate it:
 - Windows: `& "$env:PYANSYS_VENV\Scripts\Activate.ps1"`
-- Linux/macOS: `source "$PYANSYS_VENV/bin/activate"`
\ No newline at end of file
+- Linux/macOS: `source "$PYANSYS_VENV/bin/activate"`
+
+## Code style
+- **Maximum line length is 100 characters.** No generated line may exceed 100 characters. Wrap long lines by splitting arguments onto separate lines or extracting intermediate variables. After writing any file, verify with: `python -c "lines=open('path').readlines(); long=[(i+1,len(l.rstrip())) for i,l in enumerate(lines) if len(l.rstrip())>100]; print(long or 'OK')"` and fix any violations before completing the step.
```
RobPasMue: Maybe it's best to tell it to run the pre-commit hooks we have? That way we ensure everything is according to guidelines

src/ansys/geometry/core/_grpc/_services/base/body_selection.py
```diff
@@ -0,0 +1,393 @@
...
+    def get_bodies_with_z_location(self, **kwargs) -> dict:
+        """Return bodies whose Z-location falls within a range."""
+        pass
+
+    @abstractmethod
+    def get_bodies_with_color(self, **kwargs) -> dict:
+        """Return bodies matching a given color."""
+        pass
+
+    # ── Instance operations ───────────────────────────────────────────────────
```
RobPasMue: Just to make sure that this file is in line with the other ones.. maybe we should remove these comments?

tests/conftest.py
```diff
@@ -127,11 +127,15 @@ def _load_incompatible_tests(version):
...
+    import fnmatch
+
     incompatible = config.incompatible_tests
     skipped = []
 
     for item in items:
-        if item.nodeid in incompatible:
+        if item.nodeid in incompatible or any(
+            fnmatch.fnmatch(item.nodeid, pattern) for pattern in incompatible
+        ):
```
RobPasMue, line 138: Nice!

src/ansys/geometry/core/__init__.py
```diff
@@ -49,6 +49,7 @@
 )
 from ansys.geometry.core.logger import LOG
 from ansys.geometry.core.modeler import Modeler
+from ansys.geometry.core.selection_builder.selection_builder import SelectionBuilder
```
RobPasMue: Not sure this is what we want.. Users shouldn't be able to instantiate it
```suggestion
```

src/ansys/geometry/core/misc/auxiliary.py
```diff
@@ -195,7 +195,9 @@ def get_bodies_from_ids(design: "Design", body_ids: list[str]) -> list["Body"]:
     -----
     This method takes a design and body ids, and gets their corresponding ``Body`` object.
     """
-    return [body for body in __traverse_all_bodies(design) if body.id in body_ids]
+    body_id_set = set(body_ids)
+    body_map = {body.id: body for body in __traverse_all_bodies(design) if body.id in body_id_set}
+    return [body_map[bid] for bid in body_ids if bid in body_map]
```
RobPasMue: Why building the map and then returning a list based on the map? Why can't we simply do
```suggestion
    return [body for body in __traverse_all_bodies(design) if body.id in body_id_set]
```

src/ansys/geometry/core/modeler.py
```diff
@@ -722,3 +723,23 @@ def get_service_logs(
...
+        -----
+        This method is only available starting on Ansys release 27R1.
+        """
+        if self._design is None or not self._design.is_active:
+            raise GeometryRuntimeError(
+                "No active design available. Please create or open a design before "
+                "creating a selection builder."
+            )
+
+        return SelectionBuilder(self._design, self._grpc_client)
```
RobPasMue, line 757: Shouldn't this behave as the rest of the tools? We instantiate them on the constructor.. or lazy build it here at least. And if a new design gets created, we can simply drop the previous instance and build a new one.

Also, we should keep the reference to the same SelectionBuilder object for a given design, rather than building them and returning them to the user right? What I mean is using the ``self._selection_builder`` approach as we do with the other tools

src/ansys/geometry/core/selection_builder/body_selection.py
```diff
@@ -0,0 +1,1622 @@
...
+            min_percentile=min_percentile,
+            max_percentile=max_percentile,
+        )
+        bodies = get_bodies_from_ids(self._design, response["response_data"][0]["bodies"])
+        return BodySelection(self._design, self._grpc_client, bodies)
+
+    @min_backend_version(27, 1, 0)
+    def filter_bodies_surface_area_percentile(
+        self,
+        min_percentile: float,
```
RobPasMue: Instead of floats we should probably use our "Real" types

src/ansys/geometry/core/modeler.py
RobPasMue: Maybe I'm wrong!! Let me know though if that's the case!

RobPasMue (commented):
I'm assuming the implementation inside the gRPC layer is fine, and same for the body selection class .. it's pretty massive :D

I'd appreciate some examples using the body selection. Do you think we could add a section to the docs showcasing its usage?

src/ansys/geometry/core/misc/auxiliary.py
```diff
@@ -195,7 +195,8 @@ def get_bodies_from_ids(design: "Design", body_ids: list[str]) -> list["Body"]:
     -----
     This method takes a design and body ids, and gets their corresponding ``Body`` object.
     """
-    return [body for body in __traverse_all_bodies(design) if body.id in body_ids]
+    body_map = {body.id: body for body in __traverse_all_bodies(design)}
+    return [body_map[bid] for bid in body_ids if bid in body_map]
```
RobPasMue, line 207: Why did we break this into 2 statements? the dictionary seems unnecessary

src/ansys/geometry/core/modeler.py
```diff
@@ -722,3 +723,23 @@ def get_service_logs(
...
+        """Create a selection builder.
+
+        Returns
+        -------
+        SelectionBuilder
+            A selection builder object for creating custom selections.
+
+        Notes
+        -----
+        This method is only available starting on Ansys release 27R1.
```
RobPasMue, line 749: Maybe adding a min backend version decorator?

RobPasMue (commented):
LGTM - just 2 small comments. Would it be possible to include examples? I think the SelectionBuilder is going to be very useful!

src/ansys/geometry/core/modeler.py
jacobrkerstetter: Sure! It technically doesn't need one since this method doesn't contact the service, but better to stop the user before they get too far.

src/ansys/geometry/core/misc/auxiliary.py
jacobrkerstetter: The key here was preserving order for the OrderBy* selection builder methods. If there is a better way to do this let me know.

RobPasMue (dismissed):

jonahrb (approved):

tests/integration/test_body_selection.py
```diff
@@ -631,6 +631,7 @@ def test_extend_to_same_number_of_edges(modeler: Modeler):
     assert len(result.items) == 8
 
 
+@pytest.skip(reason="extend_to_same_color has a backend bug")
```
RobPasMue, line 634: I'm guessing this one will be eventually fixed, right?

RobPasMue (commented):
Thanks for merging!

tests/integration/test_body_selection.py
RobPasMue: Ah, saw it was fixed in https://github.com/ansys/pyansys-geometry/pull/2922/commits/a24cc19371c526c70f1793e1f5833826c0efb70b

PR #2965, opened 2026-07-21
feat: Added capability to create NURBS from JSON

Gonzi09:
## Description
This PR adds baseline support for creating NURBS entities from JSON payloads with optional top-level named entries.

General behavior :

Source can be a dictionary or a JSON file path.

- If the top-level object is already a valid payload, it is used directly.
- If the top-level object contains multiple named entries:
  - element_name can be provided to select one explicitly.
  - If element_name is not provided, the first valid entry is selected.
- If no valid entry is found, a ValueError is raised.
- If weights are omitted, defaults are applied as 1.0 per control point.

## Checklist


src/ansys/geometry/core/shapes/curves/nurbs.py
```diff
@@ -222,6 +224,59 @@ def fit_curve_from_points(
...
+                    )
+            else:
+                selected = None
+                for _, element in source.items():
+                    if isinstance(element, dict) and required_keys.issubset(element):
+                        selected = element
+                        break
+                if selected is None:
+                    raise ValueError("No valid NURBS curve element was found in JSON payload.")
+                source = selected
```
RobPasMue: What if we used Pydantic models for this operation? Pydantic models allow you to read from a JSON object directly and map them into a Python file. They also allow you to define some post-validation methods for their properties. I think we could simplify all this by large!

RobPasMue: The Pydantic model would also allow us to dump NURBS curves to JSON files (thus making it not only a read-wise logic but also a persistence option =) )

jacobrkerstetter: Seconding this! Great idea.

RobPasMue:
@Gonzi09 - please make sure that the tests pass before asking for a review! =)

src/ansys/geometry/core/shapes/curves/nurbs.py
```diff
@@ -222,6 +293,93 @@ def fit_curve_from_points(
...
+        return cls.from_control_points(
+            control_points=[Point3D(pt) for pt in model.control_points],
+            degree=model.degree,
+            knots=model.knots,
+            weights=model.effective_weights(),
+        )
+
+    @classmethod
+    @check_input_types
+    def from_json(
```
RobPasMue: Let's call this ``from_json_file``

src/ansys/geometry/core/shapes/curves/nurbs.py
```diff
@@ -222,6 +293,93 @@ def fit_curve_from_points(
...
+            raise ValueError(f"Element(s) {missing} were not found in JSON payload.")
+
+        built = {
+            name: cls._curve_from_model(NURBSCurveModel._validate_or_explain(name, raw[name]))
+            for name in names_to_build
+        }
+
+        if len(built) == 1:
+            return next(iter(built.values()))
+        return built
```
RobPasMue, line 350: I'd keep it consistent.. always return a dict - even if it is single key-value

src/ansys/geometry/core/shapes/curves/nurbs.py
```diff
@@ -46,6 +50,73 @@
...
+                and data["control_points"]
+                and len(data["control_points"][0]) == 2
+            ):
+                raise ValueError(
+                    f"Element '{name}' looks like a 2D NURBS sketch curve "
+                    f"(its control points have 2 coordinates, not 3). "
+                    f"Use SketchNurbs.from_json() instead of "
+                    f"NURBSCurve.from_json() for this file."
+                ) from e
+            raise
```
RobPasMue, line 116: Interesting approach!

src/ansys/geometry/core/shapes/curves/nurbs.py
```diff
@@ -222,6 +293,93 @@ def fit_curve_from_points(
...
+        ------
+        ValueError
+            If any name in ``elements`` is not found in the JSON payload,
+            or if an element's data looks like it belongs to a different
+            NURBS entity type (e.g. a surface or a sketch curve).
+        """
+        path = Path(source)
+        json_str = path.read_text(encoding="utf-8") if path.exists() else str(source)
+
+        raw = _pydantic_from_json(json_str)
```
RobPasMue: Do you need this? Or can you live with a simple json to dict conversion? I'm not sure is necessary that you perform the full pydantic model build only to extract the keys

src/ansys/geometry/core/shapes/curves/nurbs.py
```diff
@@ -222,6 +293,93 @@ def fit_curve_from_points(
...
+            The JSON string representation of this NURBS curve.
+
+        """
+        model = NURBSCurveModel(
+            degree=self.degree,
+            knots=[float(k) for k in self.knots],
+            control_points=[tuple(float(c) for c in pt) for pt in self.control_points],
+            weights=[float(w) for w in self.weights],
+        )
+        json_str = pydantic_core.to_json({element_name: model.model_dump()}).decode("utf-8")
```
RobPasMue:
```suggestion
        json_str = model.model_dump_json(indent=2)
```

Why can't we do something like this to get the JSON? Is the "element_name" necessary? Feel free to disregard!

pyproject.toml
```diff
@@ -33,6 +33,7 @@ dependencies = [
   "numpy<3,>=1.20.3",
   "pint<1,>=0.18",
   "protobuf>=3.20.2",
+  "pydantic>=2.0",
```
RobPasMue, line 36: You will also need to add a specific version to the tests target

src/ansys/geometry/core/shapes/curves/nurbs.py
jacobrkerstetter: agreed!

src/ansys/geometry/core/shapes/curves/nurbs.py
```diff
@@ -222,6 +293,93 @@ def fit_curve_from_points(
...
+        built = {
+            name: cls._curve_from_model(NURBSCurveModel._validate_or_explain(name, raw[name]))
+            for name in names_to_build
+        }
+
+        if len(built) == 1:
+            return next(iter(built.values()))
+        return built
+
+    def to_json(self, path: Union[str, Path] = None, element_name: str = "curve") -> str:
```
jacobrkerstetter: If we change the name of the other method let's make this `to_json_file`

jacobrkerstetter (dismissed):
Thanks for the added test coverage! Also, any of the general comments about NURBSCurve apply to surface + sketch.

src/ansys/geometry/core/shapes/curves/nurbs.py
Gonzi09: The idea behind "element_name" is that it work as a wrapper, by which one requires the "name" for that nurb. Otherwise, we wont be able to do a round-trip to_json_file and then from_json_file since the latter requires that "element_name". Likewise, it would give freedom to the engineers using our software to create different portions of  nurbs instead of a whole json file / geometry , allowing debugging and testing easier to achieve.

pyproject.toml
```diff
@@ -101,6 +101,7 @@ general-all = [
   "numpy==2.5.1",
   "pint==0.25.3",
   "protobuf==7.35.1",
+  "pydantic>=2.0",
```
jacobrkerstetter: @RobPasMue Let me know if this looks right! I'm curious myself how to do this 😄

src/ansys/geometry/core/sketch/nurbs.py
```diff
@@ -215,6 +286,129 @@ def fit_curve_from_points(
...
         return nurbs_curve
 
+    @classmethod
+    @check_input_types
+    def from_control_points(
+        cls,
+        control_points: list[Point2D],
+        degree: int,
+        knots: list[Real],
+        weights: list[Real] = None,
```
jacobrkerstetter: What copilot is suggesting here is correct. It should be `weights: list[Real] | None = None`

jacobrkerstetter (commented):
LGTM besides the one type hint needing fixed - let's let Roberto do one more review of the Pydantic dependency and then one of us can approve!

pyproject.toml
```diff
@@ -100,6 +101,7 @@ general-all = [
...
```
RobPasMue: On the testing side we should pin the version down to a specific version. For example, the latest one available:
```suggestion
  "pydantic==2.13.4",
```

pyproject.toml
RobPasMue: See https://github.com/ansys/pyansys-geometry/pull/2965/commits/fc26752bbcec1159715e7a99d912f1b26ae77d48 and https://github.com/ansys/pyansys-geometry/pull/2965/commits/99c65c1f79e0c5ce261e4f64fd0dfe85232bb6d3

The first one is done manually. The second commit is the result of running `` uv lock ``

src/ansys/geometry/core/shapes/curves/nurbs.py
```diff
@@ -222,6 +293,90 @@ def fit_curve_from_points(
...
+            raise ValueError(f"Element(s) {missing} were not found in JSON payload.")
+
+        built = {
+            name: cls._curve_from_model(NURBSCurveModel._validate_or_explain(name, raw[name]))
+            for name in names_to_build
+        }
+
+        return built
+
+    def to_json_file(self, path: Union[str, Path] = None, element_name: str = "curve") -> str:
```
RobPasMue: There is a conceptual issue with this method...

It is called ``to_json_file`` but it returns the JSON string... and the path to save the file is optional! Doesn't make sense... This applies in the other modules as well

RobPasMue (changes requested):

src/ansys/geometry/core/shapes/curves/nurbs.py
```diff
@@ -350,32 +349,44 @@ def from_json_file(
 
         return built
 
-    def to_json_file(self, path: Union[str, Path] = None, element_name: str = "curve") -> str:
-        """Serialize this NURBS curve to a JSON string, wrapped under a named element.
+    @classmethod
+    def to_json_file(
```
RobPasMue, line 353: I would expect this method to return the Path object after the write operation =)

src/ansys/geometry/core/shapes/surfaces/nurbs.py
```diff
@@ -443,27 +443,49 @@ def from_json_file(
...
-            weights=self.weights,
-            origin=self.origin,
-            reference=self.dir_x,
-            axis=self.dir_z,
-        )
-
-        json_str = model.model_dump_json(indent=2)
-        path = Path(destination)
+    @classmethod
+    def to_json_file(
```
RobPasMue, line 447: Same feedback

src/ansys/geometry/core/sketch/nurbs.py
```diff
@@ -382,32 +381,46 @@ def from_json_file(
 
         return built
 
-    def to_json_file(self, path: Union[str, Path] = None, element_name: str = "curve") -> str:
-        """Serialize this NURBS sketch curve to a JSON string, wrapped under a named element.
+    @classmethod
+    def to_json_file(
```
RobPasMue, line 385: Same feedback

RobPasMue (approved):

PR #3005, opened 2026-08-05
chore: pin `trame-client` to fix doc-build errors

moe-ad:
`trame-client` `3.13.3` (a transitive doc dependency) released earlier today is breaking the documentation build. This should fix it.


pyproject.toml
```diff
@@ -108,6 +108,7 @@ general-all = [
   "scipy==1.18.0",
   "semver==3.0.4",
   "six==1.17.0",
+  "trame-client!=3.13.3",
```
moe-ad:
```suggestion
  "trame-client<3.13.3",
```

moe-ad:
`trame-client` **3.13.5** has been released, which addresses the root cause (see https://github.com/Kitware/trame-client/issues/67). Closing.

PR #3023, opened 2026-08-11
feat: Added support to highlight faces with plotter

Gonzi09:
## Description
Added capability to highlight selected faces with GeometryPlotter

## Checklist


doc/source/examples/05_tools/face_selection_highlight.mystnb
```diff
@@ -0,0 +1,125 @@
...
+```{code-cell} ipython3
+def find_example_file(filename):
+    """Find an integration test file by searching parent directories."""
+    current = Path.cwd().resolve()
+    for candidate_root in [current, *current.parents]:
+        candidate = candidate_root / "tests" / "integration" / "files" / filename
+        if candidate.exists():
+            print(f"Using local file: {candidate}")
+            return candidate
+    raise FileNotFoundError(f"Could not locate {filename} in tests/integration/files")
```
jacobrkerstetter: Look at how we grab and download files in the other examples. We download them from GH

doc/source/examples/05_tools/face_selection_highlight.mystnb
```diff
@@ -0,0 +1,125 @@
...
+options.import_names = True
+
+design = modeler.open_file(file_path=file_path, import_options=options)
+design.plot()
+```
+
+## Build the face selection to highlight
+
+Create a face selection and progressively narrow it down by using the same
+filtering workflow as the local plotting script.
```
jacobrkerstetter: This comment isn't applicable to the example.
```suggestion
Create a face selection and progressively narrow it down by using a
filtering workflow.
```

doc/source/conf.py
```diff
@@ -364,6 +364,7 @@ def intersphinx_pyansys_geometry(switcher_version: str):
     "examples/04_applied/04_mechanical_named_selections": "_static/thumbnails/mechanical_ns.png",
     "examples/04_applied/05_pmdb_export_options": "_static/thumbnails/pmdb_export_options.png",
     "examples/04_applied/06_solder_ball": "_static/thumbnails/solder_ball.png",
+    "examples/05_tools/face_selection_highlight": "_static/thumbnails/measurement_tools.png",
```
jacobrkerstetter: Can we give this a custom thumbnail?

src/ansys/geometry/core/plotting/plotter.py
```diff
@@ -507,7 +547,14 @@ def plot_iter(
             _ = self.plot(object, name_filter, **plotting_options)
 
     # Override add function from plotter
-    def plot(self, plottable_object: Any, name_filter: str = None, **plotting_options) -> None:
+    def plot(
+        self,
+        plottable_object: Any,
+        name_filter: str = None,
+        highlight: FaceSelection | None = None,
```
jacobrkerstetter: Let's make this accept all types of Selections

src/ansys/geometry/core/plotting/plotter.py
```diff
@@ -516,6 +563,10 @@ def plot(self, plottable_object: Any, name_filter: str = None, **plotting_option
             Regular expression with the desired name or names you want to include in the plotter.
```
jacobrkerstetter: Not your fault - but we could change this in your PR

src/ansys/geometry/core/plotting/plotter.py
```diff
@@ -532,6 +583,9 @@ def plot(self, plottable_object: Any, name_filter: str = None, **plotting_option
             plotting_options.pop("merge_component", None)
         else:
             merge_component = None
+
+        exclude_face_ids = {face.id for face in highlight.items} if highlight is not None else None
```
jacobrkerstetter: I think we're going to need to make it more general. This would get very long very quickly if we have to do this for every selection type added in the future.

tests/integration/test_plotter.py
```diff
@@ -1590,3 +1593,130 @@ def test_plot_datum_plane(modeler: Modeler, verify_image_cache):
...
+@skip_no_xserver
+def test_show_returns_unwrapped_edge_plot():
+    """Test that show() unwraps EdgePlot into its edge_object."""
+    pl = GeometryPlotter(allow_picking=True)
+    mock_edge = MagicMock()
+    edge_plot = MagicMock(spec=EdgePlot)
+    edge_plot.edge_object = mock_edge
+    with patch.object(pl._backend, "show", return_value=[edge_plot]):
+        result = pl.show()
+    assert result == [mock_edge]
```
jacobrkerstetter: Not sure if these are necessary. Are they related to the PR or just general coverage improvement?

Gonzi09: This is testing coverage

doc/source/examples/05_tools/face_selection_highlight.mystnb
jacobrkerstetter: Look at how we do it here: https://github.com/ansys/pyansys-geometry/blob/main/doc/source/examples/05_tools/repair_tools.mystnb

src/ansys/geometry/core/plotting/plotter.py
```diff
@@ -295,7 +302,19 @@ def add_body(self, body: Body, merge: bool = False, **plotting_options: dict | N
             # Inserted in 25R2
             pass
 
-        if self.use_service_colors:
+        if exclude_ids and body.id in exclude_ids:
+            return
```
jacobrkerstetter: If the body is part of the selection we are trying to plot, would this just return without plotting it at all? I think instead of return, we should change the body color inside this if statement. Thinking towards when we want to plot a selection with say 2 bodies and 2 faces

src/ansys/geometry/core/plotting/plotter.py
```diff
@@ -388,7 +410,18 @@ def add_component(
             Keyword arguments. For allowable keyword arguments, see the
             :meth:`Plotter.add_mesh <pyvista.Plotter.add_mesh>` method.
         """
+        if exclude_ids and component.id in exclude_ids:
+            return
```
jacobrkerstetter: Same comment as above. If we have a selection that contains a component, would it be left out of the plot altogether?

src/ansys/geometry/core/plotting/plotter.py
```diff
@@ -532,6 +596,25 @@ def plot(self, plottable_object: Any, name_filter: str = None, **plotting_option
...
+        highlight_items = []
+        if highlight is not None:
+            for item in highlight.items:
+                try:
+                    if not hasattr(item, "color") or not hasattr(item, "id"):
+                        raise AttributeError
+                    item.color = highlight_color
+                    item_id = item.id
+                    highlight_items.append((item, item_id))
+                except (AttributeError, GeometryRuntimeError):
```
jacobrkerstetter, line 603: I like this implementation! Question - when would we catch a GeometryRuntimeError here? On first glance, AttributeError seems sufficient, but if it's needed feel free to keep it!

tests/integration/test_plotter.py
```diff
@@ -1590,3 +1594,195 @@ def test_plot_datum_plane(modeler: Modeler, verify_image_cache):
...
+        Sketch().box(Point2D([0, 0], UNITS.m), Quantity(2, UNITS.m), Quantity(2, UNITS.m)),
+        Quantity(2, UNITS.m),
+    )
+
+    pl = GeometryPlotter()
+    pl.plot(design, highlight=None)
+    pl.show(screenshot=Path(IMAGE_RESULTS_DIR, "test_plot_highlight_none.png"))
+
+
+FILES_DIR = Path(Path(__file__).parent, "files")
```
jacobrkerstetter: This is defined in the conftest.py - we can import from there

tests/integration/test_plotter.py
```diff
@@ -1590,3 +1594,195 @@ def test_plot_datum_plane(modeler: Modeler, verify_image_cache):
...
+    design.extrude_sketch(
+        "Box",
+        Sketch().box(Point2D([0, 0], UNITS.m), Quantity(2, UNITS.m), Quantity(2, UNITS.m)),
+        Quantity(2, UNITS.m),
+    )
+    selection = TypedSelection([design.bodies[0].edges[0]])
+
+    plotter = GeometryPlotter()
+    caplog.set_level(logging.WARNING, logger="PyAnsys_Geometry_global")
+    modeler.client.log_level = logging.WARNING
```
jacobrkerstetter, line 1789: Are both of these lines needed?

tests/integration/test_plotter.py
```diff
@@ -1590,3 +1594,195 @@ def test_plot_datum_plane(modeler: Modeler, verify_image_cache):
...
+        "Box",
+        Sketch().box(Point2D([0, 0], UNITS.m), Quantity(2, UNITS.m), Quantity(2, UNITS.m)),
+        Quantity(2, UNITS.m),
+    )
+
+    plotter = GeometryPlotter(use_service_colors=False)
+    plotter.add_body(body, exclude_ids={body.id})
+    plotter.add_component(design, merge_component=True, exclude_ids={body.id})
+    plotter.add_component(design, exclude_ids={design.id})
+    plotter.show()
```
jacobrkerstetter: This relates to my comments above. I don't think this is the behavior we want. If a body is part of a selection - we want to show it, just in a different color.

tests/integration/test_plotter.py
```diff
@@ -1590,3 +1594,195 @@ def test_plot_datum_plane(modeler: Modeler, verify_image_cache):
...
+        Sketch().box(Point2D([0, 0], UNITS.m), Quantity(2, UNITS.m), Quantity(2, UNITS.m)),
+        Quantity(2, UNITS.m),
+    )
+    selection = TypedSelection([body.faces[0]])
+
+    plotter = GeometryPlotter()
+    plotter.plot(design, highlight=selection, highlight_color="yellow")
+    plotter.show()
+
+    assert body.faces[0].color == "#ffff00ff"
```
jacobrkerstetter, line 1750: Let's just toss these lines at the end of test_plot_highlight_filtered_selection, passing in the FaceSelection (because it derives from TypedSelection). Then we can have 1 less test

doc/source/examples/05_tools/face_selection_highlight.mystnb
```diff
@@ -0,0 +1,99 @@
...
+```
+
+## Open the bracket file
+
+```{code-cell} ipython3
+file = Path("Bracket_Static_Stress.dsco")
+options = ImportOptions()
+options.import_names = True
+design = modeler.open_file(file_path=file, import_options=options)
+```
```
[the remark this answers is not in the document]
jacobrkerstetter: This is good feedback - definitely want to use launch_modeler and then the downloading is the same as I said below

src/ansys/geometry/core/plotting/plotter.py
Gonzi09: AttributeError covers items that do not expose color or id, or whose color assignment is not supported locally. ont he other hand, GeometryRuntimeError is needed as a defensie catch because these properties can be backed by remote GeometryService calls. For example, assigning item.color or reading item.id may fail if the entity was deleted, the design/server connection is no longer valid, or the backend cannot resolve the entity.

In those cases, the item is skipped and a warning is logged instead of aborting the entire plot. So I think it is useful to keep GeometryRuntimeError here.

jacobrkerstetter: Makes sense!

tests/_incompatible_tests.yml
```diff
@@ -512,6 +515,7 @@ backends:
       - tests/integration/test_design_import.py::test_importing_with_sc_colors
       # Root part CS not loaded until 27.1
       - tests/integration/test_design.py::test_coordinate_systems_on_design_refresh
+      - tests/integration/test_plotter.py::test_plot_highlight_bracket_face_selection
```
jacobrkerstetter: One final comment: could we move these excluded tests into the section with the comment "model used is too new for this version"?

jacobrkerstetter (dismissed):

jacobrkerstetter (approved):

PR #3055, opened 2026-08-19
fix: close server process on license checkout failure

jacobrkerstetter:
## Description
Close the ApiServer process if a license checkout fails. Previously, we would have an orphan process

## Issue linked
Related to TFS issue

## Checklist


src/ansys/geometry/core/connection/product_instance.py
```diff
@@ -324,7 +324,8 @@ def prepare_and_start_backend(
     ):
         # If the user has set the ANSYS_GEOMETRY_SERVICE_ROOT environment variable,
         # we will use it as the root folder for the Geometry Service.
-        pass
+        # We will still set the version to avoid later errors
+        version = get_latest_ansys_installation()[0]
```
RobPasMue: Are we sure about this change? This is specially critical for the ANSYS_GEOMETRY_SERVICE_ROOT case...

The system cannot easily determine what version of the geometry Service it is.. and thus we add all the potential licensing client env vars. If we do this, the mechanism might break unexpectedly if you are using an old release.. or you don't have any versions installed under the default dirs... I would recommend to revert this change

RobPasMue (changes requested):

src/ansys/geometry/core/connection/product_instance.py
jacobrkerstetter: Good call... the reason I put that there is because if the env var is set and there is no version provided, the next check on version will fail trying to compare None and int

RobPasMue (approved):
LGTM!

PR #3128, opened 2026-09-22
feat: length scale implementation

### LINKED ISSUE #2696: No API to set SpaceClaim Model Size (Small/Normal/Large) - causes 'modeling range exceeded' for large CFD domains
## Summary

When creating large CFD geometry (e.g., fluid domains for urban building clusters), the SpaceClaim **Model Size** setting (also known as **Length Scale**: Small / Normal / Large) limits the valid coordinate range:

| Model Size | Valid Coordinate Range |
|---|---|
| Small | ~±100 units |
| Normal | ~±1,000 units |
| Large | ~±1,000,000 units |

For a 4×3 building matrix (each 60m×20m×20m) with upstream/downstream/lateral buffers (5H/10H/5H), the fluid domain reaches ~500m+, exceeding the Normal range and causing a **"modeling range exceeded"** error.

## Problem

There is no way to set the model size through the PyAnsys Geometry API:

1. **No launch parameter** — `launch_modeler_with_spaceclaim()` has no model size option
2. **No modeler method** — No API on `Modeler` or `Design` to change it after launch
3. **Config file gets overwritten** — The `UnitsLengthScale` setting in `user.config` is reverted by SpaceClaim on startup
4. **No environment variable** — No documented env var to control initial model size

## Questions

1. Is there an existing API or mechanism to set the model size that I'm missing?
2. If not, could this be added — either as a launch parameter or a post-launch method on `Modeler`?
3. Is there a recommended approach for CFD-scale geometries that require coordinates beyond 1000 units?

## Environment

- PyAnsys Geometry: latest (PyPI)
- ANSYS version: v252
- OS: Windows 11
- Python: 3.13

## Use Case

CFD simulation geometry preparation for external aerodynamics. Building clusters of N×M matrices (e.g., 4×3, each 60m×20m) require fluid domains of 500m+, which is a standard scale for urban wind flow simulations.

jacobrkerstetter:
## Description
- Handle getting/setting length scale from imported models
- Test large and small length scale files

## Issue linked
Closes #2696 

## Checklist


src/ansys/geometry/core/modeler.py
```diff
@@ -508,7 +516,21 @@ def open_file(
...
+        if self.client.services.version == GeometryApiProtos.V1:
+            from ansys.geometry.core.designer.design import LengthScale
+
+            self._length_scale = self.client.services.designs.get_length_scale(
+                design_id=design.design_id
+            ).get("scale")
+            if self._length_scale == LengthScale.SMALL:
+                DEFAULT_UNITS.SERVER_LENGTH = UNITS.mm
+            if self._length_scale == LengthScale.LARGE:
+                DEFAULT_UNITS.SERVER_LENGTH = UNITS.km
```
RobPasMue, line 531: What if a user changes the server length halfway through the session? The default units won't be adapted, right?

jacobrkerstetter: This only handles the import case. In SC/Disco are you able to change length scale within the session? I thought that was not allowed

src/ansys/geometry/core/modeler.py
```diff
@@ -166,6 +169,11 @@ def design(self) -> "Design":
         """Retrieve the design within the modeler workspace."""
         return self._design
 
+    @property
+    def length_scale(self) -> "LengthScale":
+        """Length scale of the active design."""
+        return self._length_scale
```
RobPasMue, line 175: I have been thinking about this... what if we move the length scale information to the ``Design`` class? It makes definitely more sense than having it at the ``Modeler`` class since:

- It is associated to a specific Design
- It requires the Design to be empty to be able to set it
- The server will provide us with the LengthScale being used for a given Design

src/ansys/geometry/core/modeler.py
```diff
@@ -745,3 +767,38 @@ def create_selection_builder(self) -> SelectionBuilder:
...
+            design_id=self.design.design_id, length_scale=length_scale
+        )
+
+        if response.get("success"):
+            if self._length_scale == LengthScale.SMALL:
+                DEFAULT_UNITS.SERVER_LENGTH = UNITS.mm
+            if self._length_scale == LengthScale.LARGE:
+                DEFAULT_UNITS.SERVER_LENGTH = UNITS.km
+        else:
+            raise GeometryRuntimeError("Failed to set length scale: " + response.get("message"))
```
RobPasMue, line 802: We might have to change this approach moving forward and have it done based on the design's properties...

I'd recommend:

- Creating a method at the ``DefaultUnitsClass`` (i.e. the DEFAULT_UNITS object is of that type) that consumes the response, and changes the value based on the ``LengthScale``.
- Make sure that this method gets called at the instantiation of a Design as well so that if a user creates a new design, it gets reset to the default units (which triggers another question... what happens in SC when you have a small design, you close it, and you create a new design... does it also get assigned small units by default? does it fallback to normal units? we need to clarify this behaviour)
- Whenever the set_length_scale method is called, we call the ``DefaultUnitsClass`` method that modifies it

RobPasMue (changes requested):

src/ansys/geometry/core/modeler.py
jacobrkerstetter: I'll raise this in today's meeting. SpaceClaim has an option to set the length scale for "this document" or "all new documents" (we don't have this as something the user can change). Everything else sounds good.

src/ansys/geometry/core/modeler.py
```diff
@@ -199,6 +191,7 @@ def create_design(self, name: str) -> "Design":
 
         # Create the new design
         design = Design(name, self)
+        design._refresh_length_scale()
```
RobPasMue: Maybe add this as part of the Design ctor logic (at the very end)? :) rather than having it as a separate call that one has to call .. we might forget about it

RobPasMue: Although I understand you want to do this on Design creation (which then, the Design is empty and you can refresh easily). But you also want to read the length scale if you read/open a file... and thus, it might make more sense to read the whenever a Design object gets created.

Maybe I'm missing something - feel free to point it out! =)

src/ansys/geometry/core/misc/measurements.py
```diff
@@ -152,6 +161,23 @@ def SERVER_ANGLE(self) -> Unit:  # noqa: N802
...
+        length_scale : LengthScale
+            Length scale of the design.
+        """
+        from ansys.geometry.core.designer.design import LengthScale
+
+        if length_scale == LengthScale.SMALL:
+            self.SERVER_LENGTH = UNITS.mm
+        elif length_scale == LengthScale.STANDARD:
+            self.SERVER_LENGTH = UNITS.m
+        elif length_scale == LengthScale.LARGE:
```
RobPasMue: I would add an else condition here and fallback to the default (meters) in case it is LengthScale.UNSPECIFIED or other value

src/ansys/geometry/core/modeler.py
```diff
@@ -508,7 +508,9 @@ def open_file(
                 if temp_zip_path.exists():
                     temp_zip_path.unlink()
 
-        return self.read_existing_design()
+        design = self.read_existing_design()
```
RobPasMue: Any reasons for this change?

RobPasMue (commented):
Looking good! Just minor coments

RobPasMue (dismissed):

RobPasMue:
Maybe documenting this with an example would be a great idea! But let's do it in a separate PR

RobPasMue (approved):

